PR #4, opened 2025-07-08
Try to assemble a stage1

mudler:
now that busybox builds with our toolchain, let's try to bring everything alltogether so we can use a container from scratch from what we have built off


mudler:
OK there are issues with the latest stage of building busybox, but will pick it up from there

PR #9, opened 2025-08-01
Add kairos factory action

mauromorales:
Just a POC to see what happens if we try to kairosify uKairos

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -19,3 +19,26 @@ jobs:
     - name: Build Docker image
       run: |
         make build
+        
+    - name: Rename and push image to ttl.sh
+      run: |
+        docker tag ukairos:latest ttl.sh/ukairos:1h
+        docker push ttl.sh/ukairos:1h
+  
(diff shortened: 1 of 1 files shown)


Itxaka:
@mauromorales can the factory run a init from main? Or its it hardcoded to a tag?

mauromorales:
@Itxaka if you pass "master" as the kairos_version it will pick the Dockerfile in the main branch but my guess is that it will then crash within kairos-init cuz it's not a semver

Itxaka:
@mauromorales sorry I meant the init version to use, if we push some changes to main, do we need to cut a new version or we can set some var in the factory to use that init version (main)?

mauromorales:
@Itxaka no unfortunately that doesn't work, but tbh to make your life easier just create a Dockerfile in this repo and use `dockerfile_path` to tell the action where it is, this will override the use of the one on the kairos repo, and then you can do there your magic on which kairos-init you want to use

mudler:
Closing as we are going to release eventually to the kairos repository

PR #26, opened 2025-09-29
Add custom-compiled kernel from scratch

mudler:
Supersedes #12 

Rebasing was quite impossible, didn't wanted to overwrite the branch to preserve history. Just moved things directly into a new one.

Note: Kernel builds, but there are still some missing pieces (e.g. modules, etc)


Dockerfile
```diff
@@ -112,25 +130,6 @@ ARG PKGCONFIG_VERSION=1.8.1
 ENV PKGCONFIG_VERSION=${PKGCONFIG_VERSION}
 RUN cd /sources/downloads && wget https://distfiles.dereferenced.org/pkgconf/pkgconf-${PKGCONFIG_VERSION}.tar.xz
 
-
-## systemd patches
```
Itxaka, line 116: why was this moved?

Dockerfile
```diff
@@ -160,6 +159,67 @@ RUN cd /sources/downloads && wget https://www.netfilter.org/projects/libmnl/file
...
+ENV FTS_VERSION=${FTS_VERSION}
+RUN cd /sources/downloads && wget https://github.com/pullmoll/musl-fts/archive/v${FTS_VERSION}.tar.gz -O musl-fts-${FTS_VERSION}.tar.gz
+
+## libtool
+
+ARG LIBTOOL_VERSION=2.5.4
+ENV LIBTOOL_VERSION=${LIBTOOL_VERSION}
+RUN cd /sources/downloads && wget https://ftpmirror.gnu.org/libtool/libtool-${LIBTOOL_VERSION}.tar.xz
+
+## musl-obstack
```
Itxaka, line 210: the what? Never heard about this in my life lmao

Dockerfile
```diff
@@ -1319,16 +1637,6 @@ RUN rsync -aHAX --keep-dirlinks  /iptables/. /skeleton
 # We don't need headers
 RUN rm -rf /skeleton/usr/include
 
-FROM quay.io/luet/base:0.36.2 AS luet-base
-
-## Install kernel
-FROM alpine AS luet-kernel
-COPY --from=luet-base /usr/bin/luet /usr/bin/luet
```
Itxaka, line 1342: bye bye!!

files/kernel/files/kernel/ukairos-x86_64.config
```diff
@@ -0,0 +1,11095 @@
+#
```
Itxaka, line 1: this is not the tiny config, is it? How build is the kernel that it builds?

Dockerfile
mudler: just moved it closer to systemd

files/kernel/files/kernel/ukairos-x86_64.config
mudler: nope that's the kernel config that I've always had manually for having some good hw compatibility (at least boots on baremetals) when using Gentoo, we can go back at tiny later, but this should at least give us a head start for building on baremetals with not a super bloat

Itxaka (approved):

PR #28, opened 2025-09-29
Fix github job

Itxaka:
build on PRs but dont push it


mudler:
got it in https://github.com/kairos-io/ukairos/pull/26/commits/bfa6e2da7bd8e80e26d6e660ee58a9ff54e61014 without adding an additional step!

PR #34, opened 2025-09-30
feat: Add dracut

Itxaka:
- Added xzutils, kmod, and dracut with specified versions to enhance build capabilities.
- Improved build commands for binutils to utilize JOBS argument for parallelism.
- Ensured necessary dependencies for dracut are included in the final image.
- Add missing ldd and ldconfig scripts compatible with musl


Dockerfile
```diff
@@ -1554,6 +1580,82 @@ RUN sed -i '/^[[:space:]]*#include[[:space:]]*<linux\/if_ether\.h>/d' extensions
...
+RUN rsync -aHAX --keep-dirlinks  /libcap/. /
+
+
+COPY --from=sources-downloader /sources/downloads/kmod.tar.gz /sources/
+RUN mkdir -p /kmod
+WORKDIR /sources
+RUN tar -xf kmod.tar.gz && mv kmod-* kmod
+WORKDIR /sources/kmod
+RUN pip3 install meson ninja
+RUN meson setup buildDir --prefix=/usr --buildtype=minsize --optimization 3 -Dmanpages=false
```
mudler, line 1751: god why mesonnnn

Dockerfile
```diff
@@ -1661,6 +1763,36 @@ RUN rsync -aHAX --keep-dirlinks  /kbd/. /skeleton
 
 COPY --from=iptables /iptables /iptables
 RUN rsync -aHAX --keep-dirlinks  /iptables/. /skeleton
+
+## kmod for modprobe, insmod, lsmod, modinfo, rmmod. Draut depends on this
+COPY --from=kmod /kmod /kmod
+RUN rsync -aHAX --keep-dirlinks  /kmod/. /skeleton
+
+## fts library, dracut depends on this
```
mudler, line 2321: I think dracut deserve its own target at this point, like dracut-grub, so we keep clean for trusted-boot only builds? wdyt?

Dockerfile
Itxaka: its nice!

Dockerfile
Itxaka: yes, its a good point

files/growpart
```diff
@@ -0,0 +1,1047 @@
+#!/bin/sh
+#    Copyright (C) 2011 Canonical Ltd.
```
mudler: can't we get it from somewhere like we do with other sources instead of committing it here?

patches/gptfdisk/002_fix_include.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/gptcurses.cc b/gptcurses.cc
+index 1b18cf2..4ebfde1 100644
```
mudler: same here, are these patches ours or took somewhere?

mudler:
> This branch has conflicts that must be resolved

Yikes, rebase is needed

files/growpart
Itxaka: AFAIK growpart was created by canonical so they are the source of truth of it, we could replicate it in go or something, after all its just a script to help resize the partitions on first boot, but unfortunately we have it required by immucore for the yip plugin that expands the partitions

patches/gptfdisk/002_fix_include.patch
Itxaka: half and half. Inspired by alpine but fitted to our use case, other are just us but its probably the same patch for anyone that builds whatever is being patched under musl.

mudler: seems they are quite 1:1 https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/main/gptfdisk/fix-wrong-include.patch -> can't we just wget this in sources ?

I also say this because patches otherwise must be with the same license of the repo (which could be problematic, mixing licenses and such)

patches/lvm2/001_fix_fopen.patch
```diff
@@ -0,0 +1,67 @@
+diff --git a/lib/commands/toolcontext.c b/lib/commands/toolcontext.c
+index 60739af..8b3fdf7 100644
+--- a/lib/commands/toolcontext.c
```
mudler: This looks taken from https://gitlab.alpinelinux.org/alpine/aports/-/raw/master/main/lvm2/fix-stdio-usage.patch - better just source that during the initial phase (licensing, etc), ditto as above

patches/gptfdisk/001_fix_musl.patch
```diff
@@ -0,0 +1,58 @@
+--- a/diskio-unix.cc
```
mudler: Also this one: https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/main/gptfdisk/fix-musl.patch

mudler (approved):
❤️

PR #40, opened 2025-10-02
Rework PAM

Itxaka:
Make it simpler, drop yescrypt (doesnt work)
Drop all users before the final image
Call sysusers to create default users from scratch
Copy a decent login.defs to support shadow


Dockerfile
```diff
@@ -2125,7 +2127,7 @@ RUN mkdir -p /shadow
 WORKDIR /sources
 RUN tar -xf shadow.tar.xz && mv shadow-* shadow
 WORKDIR /sources/shadow
-RUN ./configure ${COMMON_CONFIGURE_ARGS} --sysconfdir=/etc --without-libbsd --with-bcrypt --with-yescrypt
+RUN ./configure ${COMMON_CONFIGURE_ARGS} --sysconfdir=/etc --without-libbsd --disable-static
```
mudler: COMMON_CONFIGURE_ARGS has already --disable-static

mudler:
```suggestion
RUN ./configure ${COMMON_CONFIGURE_ARGS} --sysconfdir=/etc --without-libbsd
```

PR #50, opened 2025-10-03
Try to fix multipath

Itxaka:
Add kmod support for systemd so it loads modules during initramfs, build multipath with proper lib dir

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ ENV BUILD=${BUILD_ARCH}-pc-linux-musl
 
 ### This stage is used to download the sources for the packages
 ### This runs in parallel with stage0 to improve build time since it's network-bound while stage0 is CPU-bound
-FROM alpine AS sources-downloader
+FROM alpine:${ALPINE_VERSION} AS sources-downloader
 
 # Install packages needed for downloading and patching sources
 RUN apk update && apk add wget git patch tar bash coreutils findutils
@@ -356,7 +356,7 @@ RUN cd /sources/downloads && wget -q https://gitlab.alpinelinux.org/alpine/aport
 ## busybox
 ARG BUSYBOX_VERSION=1.37.0
 ENV BUSYBOX_VERSION=${BUSYBOX_VERSION}
-RUN cd /sources/downloads && wget -q https://busybox.net/downloads/busybox-${BUSYBOX_VERSION}.tar.bz2
+RUN cd /sources/downloads && wget -q --no-check-certificate https://busybox.net/downloads/busybox-${BUSYBOX_VERSION}.tar.bz2
 
 ## musl
 ARG MUSL_VERSION=1.2.5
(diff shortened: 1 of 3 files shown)


Itxaka:
Its still failing, but at least its now loading the proper modules required by any services inthe initramfs :D

Dockerfile
```diff
@@ -1432,6 +1474,7 @@ RUN /usr/bin/meson setup buildDir \
       --buildtype=release     \
       -D dbus=true  \
       -D pam=enabled \
+      -D kmod=true \
```
Itxaka, line 1493: this is needed so the load-kernel-modules works in initramfs and services can request dependencies on modprobe@$MOD.service and systemd will insmod it automatically

Dockerfile
```diff
@@ -356,7 +356,7 @@ RUN cd /sources/downloads && wget -q https://gitlab.alpinelinux.org/alpine/aport
 ## busybox
 ARG BUSYBOX_VERSION=1.37.0
 ENV BUSYBOX_VERSION=${BUSYBOX_VERSION}
-RUN cd /sources/downloads && wget -q https://busybox.net/downloads/busybox-${BUSYBOX_VERSION}.tar.bz2
+RUN cd /sources/downloads && wget -q --no-check-certificate https://busybox.net/downloads/busybox-${BUSYBOX_VERSION}.tar.bz2
```
mudler, line 360: danke!

Dockerfile
```diff
@@ -356,7 +356,7 @@ RUN cd /sources/downloads && wget -q https://gitlab.alpinelinux.org/alpine/aport
 ## busybox
 ARG BUSYBOX_VERSION=1.37.0
 ENV BUSYBOX_VERSION=${BUSYBOX_VERSION}
```
mudler, line 358:
```suggestion
ENV BUSYBOX_VERSION=${BUSYBOX_VERSION}
# XXX: Temporary workaround as busybox currently have expired certificates
```

mudler (approved):

Dockerfile
mudler: good catch!

Itxaka:
So this makes it work, but on exit due to threads probably it coredumps. Its not a big deal as it unlinks the pid file properly I think but it looks pretty bad

Dockerfile
```diff
@@ -2402,6 +2452,9 @@ RUN rsync -aHAX --keep-dirlinks  /lvm2/. /skeleton/
 
 COPY --from=multipath-tools /multipath-tools /multipath-tools
 RUN rsync -aHAX --keep-dirlinks  /multipath-tools/. /skeleton/
+## Copy libgcc_s.so.1 for multipathd deps and gptfdisk
```
Itxaka, line 2457: instead of building them static into gptfdisk and multipathd, lets just copy the needed libs in the system

Dockerfile
```diff
@@ -2132,11 +2180,12 @@ WORKDIR /sources/gptfdisk
 RUN patch -p1 < /sources/patches/aport/main/gptfdisk/fix-musl.patch
 RUN patch -p1 < /sources/patches/aport/main/gptfdisk/fix-wrong-include.patch
 ## TODO: we need to get rid of these LDFLAGS, makes the binary around 9MB
-RUN LDFLAGS="-static-libstdc++ -static-libgcc" make -s -j${JOBS} sgdisk
+RUN make -s -j${JOBS} sgdisk
```
Itxaka, line 2183: no more need for this, maybe common args no?

Dockerfile
```diff
@@ -2053,15 +2088,28 @@ RUN rsync -aHAX --keep-dirlinks  /util-linux/. /
...
 RUN mkdir -p /multipath-tools
 WORKDIR /sources
 RUN tar -xf multipath-tools.tar.gz && mv multipath-tools-* multipath-tools
 WORKDIR /sources/multipath-tools
-# make libgcc static to avoid needing libgcc_s at runtime
-ENV CC="gcc -static-libgcc"
-RUN make -s -j${JOBS} && make -s -j${JOBS} install DESTDIR=/multipath-tools && make -s -j${JOBS} install
-
+# patch it
+RUN patch -p1 < /sources/patches/aport/main/multipath-tools/0001-Disable-O2.patch
```
Itxaka, line 2104: now patch it with alpine  patches

Dockerfile
```diff
@@ -2446,8 +2499,8 @@ RUN rsync -aHAX --keep-dirlinks  /xxhash/. /skeleton
 
 ## strace, disabled but if we need to debug this is very useful to add
 ## Just uncomment this and you will get it in the final image
-# COPY --from=strace /strace /strace
-# RUN rsync -aHAX --keep-dirlinks  /strace/. /skeleton
+COPY --from=strace /strace /strace
```
Itxaka: wops, let me remove this

Dockerfile
```diff
@@ -644,6 +648,12 @@ RUN mkdir -p /skeleton/usr/include/asm-generic && rsync -aHAX --keep-dirlinks  /
 
 COPY --from=alpine-hack /usr/include/mtd /mtd
 RUN mkdir -p /skeleton/usr/include/mtd && rsync -aHAX --keep-dirlinks  /mtd/. /skeleton/usr/include/mtd
+
+# Provide ldconfig in the image
```
Itxaka, line 652: Basically a lot of things using libtool will call ldconfig during install, I guess to bring the needed libs in case htey miss something. This provides the ldconfig that we ship from the start.

Unfortunately it should also provide readelf, but we cant get that until later in the process.

Dockerfile
```diff
@@ -720,7 +730,7 @@ ARG XXHASH_VERSION=0.8.3
 ENV XXHASH_VERSION=${XXHASH_VERSION}
 
 COPY --from=sources-downloader /sources/downloads/xxHash-${XXHASH_VERSION}.tar.gz /sources/
-
+ENV CC="gcc"
```
Itxaka, line 733: was complaining about not finding cc

Dockerfile
```diff
@@ -2009,15 +2025,34 @@ WORKDIR /sources/jsonc-build/
...
+RUN rsync -aHAX --keep-dirlinks  /pax-utils/. /
+
 COPY --from=sources-downloader /sources/downloads/urcu.tar.bz2 /sources/
-RUN mkdir -p /urcu
 WORKDIR /sources
+RUN mkdir -p /urcu
 RUN tar -xf urcu.tar.bz2 && mv userspace-rcu-* urcu
 WORKDIR /sources/urcu
-RUN ./configure ${COMMON_CONFIGURE_ARGS}
+RUN ./configure ${COMMON_CONFIGURE_ARGS} --disable-static --enable-shared --sysconfdir=/etc --mandir=/usr/share/man --infodir=/usr/share/info --localstatedir=/var
```
Itxaka, line 2055: extra flags needed.

PR #59, opened 2025-10-08
Add openiscsi to image

Itxaka:


Itxaka:
probably wont pass any tests as it will fail to enable iscsi, which IIRC its enabled on boot. Prorbably missing kernel modules, of course :D

Dockerfile
```diff
@@ -2391,6 +2399,35 @@ RUN tar -xf sudo.tar.gz && mv sudo-* sudo
...
+COPY --from=libcap /libcap /libcap
+RUN rsync -aHAX --keep-dirlinks  /libcap/. /
+
+COPY --from=sources-downloader /sources/downloads/openscsi.tar.gz /sources/
+RUN pip3 install meson ninja
+RUN mkdir -p /openscsi
+WORKDIR /sources
+RUN tar -xf openscsi.tar.gz && mv open-iscsi-* openscsi
+WORKDIR /sources/openscsi
+RUN meson setup buildDir --prefix=/usr --buildtype=minsize --optimization 3 -D isns=disabled
```
mudler, line 2435: does this install also the systemd services?

mudler (approved):
!!! 💯

Dockerfile
Itxaka: Yes!

Itxaka:
now with missing modules which fix the basic iscsi services :D

PR #60, opened 2025-10-08
feat: Optimize Dockerfile for smaller image size

Itxaka:
- Added `install-strip` to various make commands to reduce the final image size by stripping unnecessary symbols and debug information.
- Updated meson build configurations to use `-Dstrip=true` for minimized binaries.
- Removed unnecessary files and directories post-installation to further decrease image size.

These changes aim to create a more efficient and lightweight Docker image, improving deployment speed and resource usage.


Dockerfile
```diff
@@ -2702,8 +2710,28 @@ COPY --from=full-image-pre-systemd /skeleton /
...
+# Remove man files, 4,9Mb
+RUN rm -rf /usr/share/man
+RUN rm -rf /usr/local/share/man
+# Remove docs 4,9Mb
+RUN rm -rf /usr/share/doc
+# remove info 3,8Mb
+RUN rm -rf /usr/share/info
+RUN rm -rf /usr/share/local/info
+# Remove static libs
+RUN find / -name '*.a' -delete
```
mudler, line 2770: just wondering if this could be potentially a breaking one? are we sure nothing is using the static libs?

mudler (approved):
naiss!

Dockerfile
Itxaka: We dont, no. But trial and error shows that curently nothing is using It and its just a line tondrop if we find something uses it

Itxaka:
:cry: seems to break stuff especially if we strip gcc libs :sob: 

Sucks becuase it went from 20Mb to 2,3...

Itxaka:
ok, reworked to not strip stuff from gcc as it seems like anything can use it. We dont save as much but still, we got 15Mb in there just in case.

Another option would be to strip gcc AND build sgdisk with static which makes sgdisk bigger but gcc much smaller

PR #74, opened 2025-10-30
feat: Support building trusted boot images

Itxaka:
- Introduced libffi version 3.5.2 to the build process so python can be built with ctypes
- Build systemd-boot in a separate process
- Allow building for systemd bootloader with BOOTLOADER var in the makefile
- Finish the systemd-only target which provides a hadron image with no dracut or grub

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -465,6 +465,10 @@ RUN cd /sources/downloads && wget -q https://github.com/open-iscsi/open-iscsi/ar
 ARG GDB_VERSION=16.3
 RUN cd /sources/downloads && wget -q https://sourceware.org/pub/gdb/releases/gdb-${GDB_VERSION}.tar.gz && mv gdb-${GDB_VERSION}.tar.gz gdb.tar.gz
 
+
+ARG LIBFFI_VERSION=3.5.2
+RUN cd /sources/downloads && wget -q https://github.com/libffi/libffi/releases/download/v${LIBFFI_VERSION}/libffi-${LIBFFI_VERSION}.tar.gz && mv libffi-${LIBFFI_VERSION}.tar.gz libffi.tar.gz
+
 FROM stage0 AS skeleton
 
 COPY ./setup_rootfs.sh ./setup_rootfs.sh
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1500,7 +1516,7 @@ RUN DESTDIR=/kmod ninja -C buildDir install && ninja -C buildDir install
 
 
 ## systemd
-FROM rsync AS systemd
+FROM rsync AS systemd-base
```
Itxaka, line 1519: base target for both systemd builds

Dockerfile
```diff
@@ -1585,11 +1607,31 @@ RUN /usr/bin/meson setup buildDir \
...
       -D docdir=/usr/share/doc/systemd-${SYSTEMD_VERSION}
 RUN ninja -C buildDir
 RUN DESTDIR=/systemd ninja -C buildDir install
 RUN ninja -C buildDir install
 
+
+FROM systemd-base AS systemd-bootloader
+ARG VERSION
+WORKDIR /sources/systemd
+ENV CFLAGS="$CFLAGS -D __UAPI_DEF_ETHHDR=0 -D _LARGEFILE64_SOURCE -D__DEFINED_wchar_t"
```
Itxaka, line 1620: this overrides the CFLAGS to be able to build systemd-boot but on the other side it will break the normal systemd-build unless we add custom patches. Seems simpler to have 2 different targets as building the efi files doesnt take long.

Makefile
```diff
@@ -4,12 +4,14 @@ AURORA_IMAGE ?= quay.io/kairos/auroraboot:v0.9.0
 TARGET ?= default
 JOBS ?= 24
 GIT_VERSION := v0.0.0-$(shell git describe --tags --always --dirty)
+BOOTLOADER ?= grub
```
Itxaka, line 7: allows overriding the bootloader to produce a different image

jimmykarily (approved):

Dockerfile
mudler: sounds good to me! maybe just add a comment so we don't forget later on why we have the dup? 🤣

mudler (approved):

PR #77, opened 2025-11-07
feat: Update Dockerfile and configuration for TPM support

Itxaka:
- Enable securityfs in kernel
- Enable TPM support in kernel
- Enable fat, exfat and ntfs configs in kernel
- Set proper charsets for FAT/EXFAT fs
- Enable crypto cyphers in kernel
- Drop pam config that its not built (lastlog2)
- Build TPM2 libs
- Build lvm2 with udev and systemd support
- Build systemd with TPM and cryptsetup support
- Workaround to build missing tpm2 related services under systemd without building the bootloader, which doesnt build with musl
- Reorder things to make deps work
- Add missing libs to full image (kmod,lzma, tss)
- Fix version passing to Makefile
- Add workflow for building trusted images in ci
- Add tests for trusted images in ci


mudler (approved):

Itxaka:
Requires a small changes in the uki auroraboot path, but merging so the cache gets filled

PR #79, opened 2025-11-11
feat: Add libxml2 support to Dockerfile and fix toolchain

Itxaka:
- Introduced libxml2 version 2.14.5 to the build process
- Added necessary steps to download, extract, and install libxml2
- Enhanced toolchain capabilities by including libxml2 for XML processing
- Fix toolchain deps. Some were missing
- Fix toolchain entrypoint. No shell was in there so it could not be used to build stuff with it as common commands would fail


jimmykarily:
Do we keep track somewhere why we introduce each dependency? E.g. where is libxml2 needed? This is extremely useful information when we try to cleanup old things, knowing why they were introduced in the first place.

jimmykarily (approved):

Itxaka:
> Do we keep track somewhere why we introduce each dependency? E.g. where is libxml2 needed? This is extremely useful information when we try to cleanup old things, knowing why they were introduced in the first place.

not really no, but the thing is, we can introduce as many deps as we want, as long as they are not added to the final image, we will just have them ready in the toolchain target to build future stuff.

This was me just having a look around and seeing that some of the more complex packages migth require the libxml and wanted to try the toolchain target to see if it worked :D

And after adding it, it just made sense to leave it in there in case we need it in the future

PR #82, opened 2025-11-12
feat: Enhance Dockerfile for small footprint

Itxaka:
- Provides a minimal busybox config for the final target
- Drop parted from final image
- Drop dosfstools from final image (provided by busybox now)
- Drop gptfdisk from final image
- Drop popt(gptfdisk dep) from final image
- Remove python artifacts
- Strip binaries safely
- Add a new target that runs all binaries in the image and verifies that they do not segfault
- Improved installation process for gptfdisk by building it with libstdc statically:

This results in the gdisk binary growing from 500Kb to 9Mb, but makes us avoid having to ship the full libstdc++ lib which is about 20Mb. Previous results to try to strip the lib produced panics so its the only way to reduce the global size.

For now only sgdisk needs that lib.

Build artifacts size:
```
hadron-init                 latest          4b7c1e8b229b   14 hours ago    173MB
hadron-static_gptdisk       latest          3776b341e27f   14 hours ago    135MB
```


Itxaka:
expected to fail until a new init version is released which drops the requirements of parted/growpart from immucore

Itxaka:
This should now pass

mudler (approved):
Nice!

mudler (approved):

PR #85, opened 2025-11-24
feat: baremetal support + cloud and default kernel configs

Itxaka:
- Move config to default
- Add a fragment to enable more baremetal drivers
- Add new targets for the 2 different kernels, virtual and baremetal
- Support building any of those by Makefile, default to baremetal
- Add firmware utility to build, pack, push, and build sysext with
  firmware


mudler (approved):
perfect!

PR #90, opened 2025-11-26
docs: Architecture

mauromorales:
I created a few architecture files with AI, it supposedly read the Docker and makefile so I hope it's correct but I'll need you to validate pls (sorry about that, but I think this is faster than if we write it ourselves atm).

They are written trying to mimic our kairos voice, but with focus on SEO

I went through them and they seem logical for the most part, but again, please have a look and see if there's any weird claim so we can fix it.


docs/content/docs/architecture/_index.md
```diff
@@ -1,9 +1,136 @@
...
+title: "Architecture Overview"
 linkTitle: "Architecture"
 icon: fa-regular fa-diagram-project
 weight: 4
 description: |
-    Hadron's desing and implementation is explained in this section
+    Understanding Hadron's design and implementation
 ---
 
+Hadron is a minimal Linux distribution built completely from scratch using a container-based build system. Unlike distributions derived from Ubuntu, Debian, or Alpine, Hadron is constructed from upstream sources with no downstream patches or modifications.
```
Itxaka: mmhhheeeeemmmh

We do carry some patches though, but those are for building under musl only. No opinionated patches or new features added by patches or behaviour changes. Not sure if we can claim this when doing patches just for fixing things like headers and hardcoded glibc functions and so on?

docs/content/docs/architecture/_index.md
```diff
@@ -1,9 +1,136 @@
...
+- **musl**: Small, secure, reproducible C library ideal for cloud-native workloads
+- **systemd**: Modern service management, observability, and ecosystem compatibility
+
+This combination delivers a small footprint with modern system management capabilities.
+
+## Architecture Components
+
+### Build System
+
+Hadron uses a multi-stage Docker build process that constructs the entire operating system from source. The build follows a Linux From Scratch (LFS) approach with modern containerization.
```
Itxaka: noice, I like the shout to LFS

docs/content/docs/architecture/_index.md
```diff
@@ -1,9 +1,136 @@
...
+
+The kernel is built from upstream sources with minimal configuration, allowing customization per architecture while maintaining a small footprint.
+
+See the [Core Components](/docs/architecture/core-components/) documentation for detailed information.
+
+### Boot Architecture
+
+Hadron supports two bootloader configurations:
+
+- **GRUB**: Traditional bootloader with BIOS and UEFI support, used with dracut for initramfs generation
```
Itxaka: Dont talk about bootloaders, talk about BIOS and UEFI if possible I think. USers dont care about bootloader but care about features (bios for all machines, UEFI for secureboot) Maybe not even bios/uefi, just trusted boot vs not trusted?

docs/content/docs/architecture/_index.md
```diff
@@ -1,9 +1,136 @@
...
+```bash
+docker pull ghcr.io/kairos-io/hadron:main 
+docker run -it --rm ghcr.io/kairos-io/hadron:main /bin/sh
+```
+
+## Security Features
+
+### Minimal Attack Surface
+
+By including only essential components, Hadron reduces the attack surface:
```
Itxaka, line 117: Maybe we can say that becuase we are so close to source, we can easily update to patch newer CVES that are fixed upstream as we dont need to carr ythose patches and do QA  over themand get them into our machinery

docs/content/docs/architecture/_index.md
mudler: we need to drop the patches bit. we do heavily patching around systemd (and it's hard to agree on what to consider as a limit here)

docs/content/docs/architecture/_index.md
```diff
@@ -1,9 +1,136 @@
...
+- Upstream sources minimize supply chain risks
+
+### Trusted Boot Support
+
+With systemd-boot, Hadron supports Secure Boot and TPM-based measurements for trusted boot scenarios.
+
+### Default Hardening
+
+The build system applies security hardening by default, including stripped binaries and minimal setuid binaries.
+
```
mudler, line 131: we could likely expand on why the choice of components, e.g. why musl and not let's say glibc? why systemd and not runit or OpenRC? (see the slides for content)

docs/content/docs/architecture/_index.md
mauromorales: Removed the patches part

docs/content/docs/architecture/boot.md
```diff
@@ -0,0 +1,216 @@
...
+
+- Linux kernel (vmlinuz)
+- Initramfs (if needed)
+- systemd stub
+- Boot configuration
+- Certificates and signatures
+
+This single EFI executable simplifies the boot process and enables trusted boot scenarios.
+
+## Boot Process Comparison
```
Itxaka: not sure if this part needs to be here, too much info? does customer cares about this

docs/content/docs/architecture/boot.md
```diff
@@ -0,0 +1,216 @@
...
+### Unified Kernel Images (UKI)
+
+Trusted Boot uses Unified Kernel Images that combine kernel, initramfs, and boot configuration into a single EFI executable:
+
+- Single file contains all boot components
+- Easier to sign and verify
+- Simpler boot process
+- Better integration with Secure Boot
+
+### Trusted Boot Capabilities
```
Itxaka, line 98: should make sure to emphasize that the user is in charge of the keys, not a third party like microsoft, fedora or ubuntu.

docs/content/docs/architecture/build-system.md
```diff
@@ -0,0 +1,174 @@
...
+
+- GCC compiler and runtime libraries
+- musl libc
+- binutils (assembler, linker, etc.)
+- make and other build utilities
+- BusyBox for basic utilities during build
+
+This environment is used to compile all subsequent packages in a consistent, reproducible manner.
+
+### Stage 1.5: Package Compilation
```
Itxaka: do we need to be this specific witht he build process? Whos the target for this?

Itxaka: plus this could chqange at any time or even be incorrect

docs/content/docs/architecture/build-system.md
```diff
@@ -0,0 +1,174 @@
...
+
+**Extend Base Image**
+
+Use the base container image as a starting point for your own Dockerfile, adding packages and configurations as needed.
+
+## Build Reproducibility
+
+The entire build process is containerized, ensuring:
+
+- **Consistent builds**: Same inputs produce identical outputs
```
Itxaka: mhhh. I would not call this reproducible as it has other meanings. We will not get the same sha of the artifact ever due to at least timestamps for example.

docs/content/docs/architecture/core-components.md
```diff
@@ -0,0 +1,273 @@
...
+C standard libraries are wrappers around the system calls of the Linux kernel, providing the interface between applications and the operating system. Hadron chose musl over glibc for several critical reasons:
+
+**Small Footprint**
+
+musl is significantly smaller than glibc, resulting in:
+
+- Smaller binary sizes
+- Reduced memory usage
+- Faster startup times
+- Smaller container images (Hadron images are ~8MB vs typical glibc-based distributions)
```
Itxaka: wat

docs/content/docs/architecture/core-components.md
```diff
@@ -0,0 +1,273 @@
...
+- **Ecosystem integration**: Less support from cloud-native tools and platforms
+- **Security features**: Fewer built-in security and resource management features
+
+For a distribution targeting cloud-native and edge workloads, systemd's modern capabilities outweigh the additional complexity, especially when combined with musl's minimal footprint.
+
+### systemd Features in Hadron
+
+The build process compiles systemd with musl support, enabling features like:
+
+- **systemd-cryptsetup**: Encrypted partition management
```
Itxaka: are we selling systemd here? Too much IMHO, I would just say that systemd is an industry standard due to the broad capabilities that it provides vs other init systems.

docs/content/docs/architecture/boot.md
mauromorales: removed it

docs/content/docs/architecture/boot.md
mauromorales: added an alert mentioning this

docs/content/docs/architecture/build-system.md
mauromorales: yup makes sense, removed it

docs/content/docs/architecture/core-components.md
mauromorales: LOLOLOL sorry about that, but I was not sure about the final size so I placed something that was ridiculous small to remember to fix, but of course I forgot

docs/content/docs/architecture/core-components.md
mauromorales: the idea was that each of these are significantly more modern than their counterparts in openRC, but indeed maybe it's a bit too much, condensed it into a single paragraph

docs/content/docs/architecture/build-system.md
mauromorales: rephrased it to keep it a bit more loose, let me know what you think

docs/content/docs/architecture/core-components.md
mauromorales: actually makes no sense to talk about sizes, removed it

docs/content/docs/architecture/_index.md
```diff
@@ -1,10 +1,137 @@
...
+- **musl**: Small, secure, reproducible C library ideal for cloud-native workloads
+- **systemd**: Modern service management, observability, and ecosystem compatibility
+
+This combination delivers a small footprint with modern system management capabilities.
+
+## Architecture Components
+
+### Build System
+
+Hadron uses a multi-stage Docker build process that constructs the entire operating system from source. The build follows a [Linux From Scratch (LFS)](https://www.linuxfromscratch.org/) approach with modern containerization.
```
mudler, line 44:
```suggestion
Hadron uses a multi-stage Docker build process that constructs the entire operating system from source. The build follows a process akin to [Linux From Scratch (LFS)](https://www.linuxfromscratch.org/) but built entirely in a single Dockerfile.
```

docs/content/docs/architecture/boot.md
```diff
@@ -0,0 +1,199 @@
...
+- Protection against boot-time attacks
+
+### Unified Kernel Images (UKI)
+
+Trusted Boot uses Unified Kernel Images that combine kernel, initramfs, and boot configuration into a single EFI executable:
+
+- Single file contains all boot components
+- Easier to sign and verify
+- Simpler boot process
+- Better integration with Secure Boot
```
mudler, line 96:
```suggestion
- Better integration with Secure Boot
- The UKI file is the whole OS, there is no switch-root to additional partitions or images
```

docs/content/docs/architecture/core-components.md
```diff
@@ -0,0 +1,265 @@
...
+These three components work together to provide a complete Linux system:
+
+**musl** provides the system call interface and standard library, enabling applications to interact with the kernel.
+
+**systemd** manages the boot process, services, and system resources, providing a modern foundation for running workloads.
+
+**Linux kernel** provides hardware abstraction, process management, and system services.
+
+Together, they form a minimal but complete operating system suitable for cloud-native and edge deployments.
+
```
mudler, line 249: what probably would make sense to add is firmwares and how users should handle it

docs/layouts/index.html
```diff
@@ -75,7 +75,7 @@ <h1>The foundation for image-based systems.</h1>
                 </p>
 
                 <div class="ctas" role="navigation" aria-label="Primary">
-                    <!-- <a href="/docs/" class="btn primary" id="btn-start">Documentation</a> -->
+                    <a href="/docs/" class="btn primary" id="btn-start">Documentation</a>
```
mudler, line 78: are we sure we want to make the documentation of hadron separated? or this is temporary as soon as we merge kairos and hadron artifacts?

mudler: I just say this because it can be really confusing: at the moment we have the quickstart on kairos.io, but we would have "some docs" on hadron-linux.io

docs/content/docs/architecture/boot.md
```diff
@@ -0,0 +1,199 @@
...
+dracut integrates with systemd in the initramfs:
+
+- Uses systemd as the init process
+- Leverages systemd's service management
+- Provides systemd units for early boot
+- Enables systemd features in initramfs
+
+### Implementation Details
+
+Standard Boot uses GRUB as the bootloader. The build system produces BIOS images for legacy hardware platforms.
```
Itxaka, line 74: I wonder if we should extend this to also mention that we support plain UEFI with no secureboot/trusted boot?

So we ship grub for bios AND efi and create hybrid isos that boot on both. The current text seem to imply that for bios you boot with grub and for EFI we boot with UKI adn thats it.

Maybe we should clarify BIOS/EFI no secureboot and Trusted Boot ?

docs/content/docs/architecture/build-system.md
```diff
@@ -0,0 +1,96 @@
...
+## Build Customization
+
+The build system is designed for customization. You can:
+
+**Modify Package Versions**
+
+Use build arguments to specify package versions:
+
+```dockerfile
+ARG KERNEL_VERSION=6.16.7
```
Itxaka, line 40: ? is this supported? I mena on some causes it may work but for example kernel or systemd are very specific on their toggles and configs, so maybe we should drop this

docs/content/docs/architecture/build-system.md
```diff
@@ -0,0 +1,96 @@
...
+ARG SYSTEMD_VERSION=257.8
+```
+
+**Add Custom Packages**
+
+Create additional build stages to include custom software:
+
+```dockerfile
+FROM stage1 AS custom-package
+COPY --from=sources-downloader /sources/custom-package.tar.gz /sources/
```
Itxaka, line 50: incrust the examples/*

docs/content/docs/architecture/build-system.md
```diff
@@ -0,0 +1,96 @@
...
+# ... build and install custom package
+```
+
+**Adjust Kernel Configuration**
+
+Modify kernel configuration files per architecture to enable or disable features.
```
Itxaka, line 56: ugh, no, lets keep this out until we figure it out. Ideally you could ship a fragment and we could apply it during build but we are not there yet

docs/content/docs/architecture/core-components.md
```diff
@@ -0,0 +1,265 @@
...
+### systemd Features in Hadron
+
+The build process compiles systemd with musl support, enabling systemd features without requiring glibc, delivering modern system management capabilities with a minimal footprint.
+
+### Build Considerations
+
+systemd is built with musl support, which requires careful configuration. The build system:
+
+- Applies necessary compatibility patches from upstream
```
Itxaka, line 152: no more patches!

docs/content/docs/architecture/core-components.md
```diff
@@ -0,0 +1,265 @@
...
+### Kernel Customization
+
+The kernel configuration can be customized per architecture:
+
+- Enable or disable specific features
+- Add hardware support
+- Configure security options
+- Optimize for specific use cases
+
+Configuration files are stored in `files/kernel/` and can be modified to suit your needs.
```
Itxaka, line 210: again, not there yet :)

docs/content/docs/architecture/core-components.md
Itxaka: do we even know at this point how its going to look like?

docs/layouts/index.html
mauromorales: @mudler If we stay distro-agnostic, Kairos cannot be the Linux distro, it has to be the abstraction layer that turns different Linux bases into an immutable, bootable system.

Hadron is then our reference distro, just like Ubuntu, openSUSE, or Alpine were before, except now we also own the base OS.

Because of that, Hadron needs its own OS-level docs and releases, kernels, CVEs, secure boot, images, so those don’t get mixed up with Kairos bugs or Kairos lifecycle releases.

At the same time, it’s totally fine (and even good) for the Hadron quickstart to live on kairos.io, just like we previously had Ubuntu quickstarts there. Kairos remains the entry point, Hadron remains the base.

PR #92, opened 2025-11-26
fix: Bump multipath-tools to 0.11.3

Itxaka:
Brings fixes for building under musl.
Drop the patches as they are no longer needed for a clean build This also fixes the coredump on multipathd exit


mudler (approved):

mudler:
love when it's about cleanup!

PR #95, opened 2025-11-27
fix: Update image tags for consistency and clarity

Itxaka:
- Changed image tags in `image_publish.yml` to use `hadron-toolchain` and `hadron-trusted` for better clarity.
- Updated Makefile to reflect the new image naming convention.
- Ensured that the build dependencies are correctly referenced for the new image names.

### DIFF
--- a/.github/workflows/image_publish.yml
+++ b/.github/workflows/image_publish.yml
@@ -71,7 +71,6 @@ jobs:
           context: ./
           file: ./Dockerfile
           platforms: linux/amd64
-          push: true
           tags: ttl.sh/hadron-toolchain:${{ github.sha }}
           labels: ${{ steps.meta.outputs.labels }}
           target: toolchain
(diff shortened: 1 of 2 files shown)


.github/workflows/image_publish.yml
```diff
@@ -72,7 +72,7 @@ jobs:
           file: ./Dockerfile
           platforms: linux/amd64
           push: true
-          tags: ttl.sh/hadron-container:${{ github.sha }}
+          tags: ttl.sh/hadron-toolchain:${{ github.sha }}
```
Itxaka, line 74: also fixed this. we are only using it for PRs but still, it should have the proper name. Im not sure why we are pushing this image anyway?

Itxaka: its not like we are using it anywhere, maybe we should drop the push on PRs...

mauromorales (approved):
I like the distinction between trusted and bios 👏 

we still need to educate our users but it's much better than asking them to choose the bootloader and getting a different "product"

mauromorales (approved):
even better :)

PR #99, opened 2025-11-27
fix: Update image names for clarity and consistency

### LINKED ISSUE #3789: Simplify Image Naming Scheme and Finalize Kubernetes-Ready Build Targets
### Summary
We need to streamline the image naming convention to reduce noise and complexity, especially for the core `hadron` build flavors. The current naming includes redundant terms like `default`, which adds unnecessary verbosity without functional benefit.

### Current Naming (Problems)
- `hadron-default` → redundant; `hadron` is already the default.
- `hadron-bios` → misleading; BIOS is implied in `hadron` unless explicitly overridden (e.g., `hadron-trusted`).
- `hadron-cloud` → no explicit boot mode, but we want to avoid cluttering names with technical details unless required.

### Proposed Final Naming Scheme
| Current Name | New Name | Rationale |
|--------------|----------|-----------|
| `hadron-default` | `hadron` | Default is implicit; no need to state it. |
| `hadron-bios` | `hadron` | BIOS is the default boot mode; no need to specify unless overridden. |
| `hadron-trusted` | `hadron-trusted` | Clear, meaningful, and consistent. |
| `hadron-cloud` | `hadron-cloud` | Lightweight, no boot mode specified (assumed minimal boot). |
| `hadron-cloud-trusted` | `hadron-cloud-trusted` | Same rationale as above; trusted variant. |
| `hadron-container` | `hadron-container` | Acceptable for debugging/vanity; keep as-is. |

### Action Items
1. **Remove `default` and `bios` from all image names** — they are redundant.
2. **Update all CI/CD pipelines** to reflect the new naming (e.g., `hadron` → `hadron-bios` only if explicitly needed).
3. **Finalize build targets for Kairos**:
   - `hadron`
   - `hadron-trusted`
   - `hadron-cloud`
   - `hadron-cloud-trusted`
4. **Document** the new naming convention in the `CONTRIBUTING.md` and `README.md`.

### Why This Matters
- Simplicity > verbosity.
- Reduces cognitive load for developers.
- Prevents confusion in scripts, tags, and image references.
- Aligns with real-world usage: `hadron-trusted` ≠ `hadron` because it changes behavior (bootloader, secure boot), not just flavor.

### Next Steps
- [ ] Update naming logic in build scripts
- [ ] Regenerate and push new images with clean names
- [ ] Update documentation
- [ ] Notify team and CI/CD maintainers

Itxaka:
- Changed container names in `image_publish.yml` to include `-container` suffix for better identification.
- Drop the bios appendix from default images
- Updated `Makefile` to reflect changes in image naming conventions.
- Rework makefile to produce BIOS images by default

Fixes: https://github.com/kairos-io/kairos/issues/3789

### DIFF
--- a/.github/workflows/image_publish.yml
+++ b/.github/workflows/image_publish.yml
@@ -106,7 +106,7 @@ jobs:
         uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051
         with:
           images: |
-            ghcr.io/${{ github.repository }}
+            ghcr.io/${{ github.repository }}-container
           tags: |
             type=semver,pattern={{raw}}
             type=ref,event=branch
@@ -136,7 +136,7 @@ jobs:
           context: ./
           file: ./Dockerfile
           platforms: linux/amd64
-          tags: ttl.sh/hadron:${{ github.sha }}
+          tags: ttl.sh/hadron-container:${{ github.sha }}
           labels: ${{ steps.meta.outputs.labels }}
           target: container
           cache-from: type=gha,scope=hadron/ci-image-main
@@ -183,7 +183,7 @@ jobs:
         uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051
         with:
           images: |
-            ghcr.io/${{ github.repository }}${{ matrix.kernel == 'cloud' && '-cloud' || '' }}-bios
+            ghcr.io/${{ github.repository }}${{ matrix.kernel == 'cloud' && '-cloud' || '' }}
           tags: |
             type=semver,pattern={{raw}}
             type=ref,event=branch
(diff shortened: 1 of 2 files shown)


mudler (approved):

mauromorales (changes requested):
I don't think this is good in terms of UX. I obviously cannot veto 😂 but I really think hadron alone should be the container or alternatively let's not have hadron alone, but let all of them be specific. We need to have a similar experience than other OSes imo. It's also less explaining and easier to understand. Last but not least even on the kairos side it's clearer, why would I use a hadron-bios if I'm later on passing trusted=true. On the other hand if I use hadron and trusted=true doesn't clearly say: hey man that's not ok

Itxaka:
>  Last but not least even on the kairos side it's clearer, why would I use a hadron-bios if I'm later on passing trusted=true. On the other hand if I use hadron and trusted=true doesn't clearly say: hey man that's not ok

I can only answer for this and is if you pass a bios onto init and tell it to build a trusted boot, then it will fail as expected. Kind of if you pass a random distro that doesnt have systemd packages available and whatnot. It is what it is, its the constraint of not having a package manager, we have to publish 2 images for this, no other way. If we were consuming something from a distro that also had its quirks we would also adapt to it, I dont see this as a weird thing or anything.

The rest, I cannot say.

PR #103, opened 2025-12-01
add some extra as examples

Itxaka:


mauromorales (approved):
noice! and LOL about the size of those things

examples/add-packages/Dockerfile.git
```diff
@@ -0,0 +1,20 @@
+# This dockerfile builds git from source using the Hadron toolchain image.
```
mudler, line 1: maybe move these under `examples/add-package` ? this way it's hard to tell if it's really used by the code or not

.github/workflows/extra.yml
```diff
@@ -0,0 +1,25 @@
+name: Build extras
```
mudler:
```suggestion
name: Build examples
```

examples/add-packages/Dockerfile.git
```diff
@@ -0,0 +1,20 @@
...
+
+FROM ghcr.io/kairos-io/hadron-toolchain:main AS builder
```
mudler, line 3: can we also have a readme? I was thinking that this example would be perfect to show how to include the package in both ways (sysext, package)

examples/add-packages/Dockerfile.doom
```diff
@@ -0,0 +1,405 @@
+# This dockerfile builds cage and all its dependencies from source using
+# a multi-stage build process. Each dependency is built in its own stage,
+# and the final stage copies the built artifacts into the final image.
+# This can be used to launch a kiosk environment using cage as the Wayland compositor.
+
+
+FROM ghcr.io/kairos-io/hadron-toolchain:main AS wayland
```
mudler, line 7: I love it!!!

what next? sway? :D

examples/add-packages/Dockerfile.git
Itxaka: thats for the docs no?
Ideally the docs can point to the examples or even incrust them int he docs and explain things better in there

examples/add-packages/Dockerfile.doom
mauromorales: no, the cool kids are now using Hyprland

mudler (approved):

PR #108, opened 2025-12-02
temporarily hide header 

mauromorales:
we don't have docs shared here atm so let's remove to have better visibility

Let me know if you also want to remove the footer @mudler


mudler (approved):
thanks!

PR #110, opened 2025-12-03
Allow the user to also build standard images with the Makefile

jimmykarily:
with:

 KUBERNETES_DISTRO=k3s make build


mudler:
this should be part of the kairos repository. I think @Itxaka is already looking into moving it there

Itxaka:
Yes, indeed, it made sense to have the init here when we could not test it as part of the Kairos workflow, as we needed everything self contained. Still I think Hadron having its own init to test it directly here is useful. This means its self contained. 

But ideally we would just reuse the Kairos Dockerfile here + factory action to build the images so the workflow its the same, but we dont need to duplicate things.

Not sure if we want to add a small test here for k3s/k0s to make sure any changes do not break at least the basic of a empty k3s deployment or something. I think this is good to have locally at least for testing ourselves.


EDIT: Not sure if what I wrote is understandable.

Basically:
 - dockerfile.init should go, we can and should use the Kairos one
 - makefile for building k8s images is good, we want to add that to our tests to know if we broke basic stuff and for that we need to be able to build the image directly here. Users can also use the makefile if they wish but that should not be its main purpouse, nor should we release core/standard images from here but from Kairos repo instead.

Dockerfile.init
```diff
@@ -7,7 +7,13 @@ FROM quay.io/kairos/kairos-init:${KAIROS_INIT} AS kairos-init
 FROM ${BASE_IMAGE} AS base-kairos
 ARG VERSION
 ARG TRUSTED_BOOT
+ARG KUBERNETES_DISTRO
```
Itxaka: We should jsut reuse the https://github.com/kairos-io/kairos/blob/master/images/Dockerfile here, not sure if downloading it the first time or something like that in the makefile as to not diverge too much. after all  thats the real dockerfiel we will be using for building.

Makefile
```diff
@@ -93,11 +96,22 @@ build-kairos:
...
+		echo "Building standard image with Kubernetes distribution: ${KUBERNETES_DISTRO}"; \
+		docker build --no-cache ${PROGRESS_FLAG} -t ${INIT_IMAGE_NAME} \
+		-f Dockerfile.init \
+		--build-arg BASE_IMAGE=${IMAGE_NAME} \
+		--build-arg TRUSTED_BOOT=$$TRUSTED_BOOT \
+		--build-arg VERSION=${VERSION} \
+		--build-arg KUBERNETES_DISTRO=${KUBERNETES_DISTRO} .; \
+	else \
+		echo "Building core image (no Kubernetes distribution)"; \
+		docker build --no-cache ${PROGRESS_FLAG} -t ${INIT_IMAGE_NAME} \
```
Itxaka: why no-cache?

Makefile
```diff
@@ -93,11 +96,22 @@ build-kairos:
...
-	--build-arg TRUSTED_BOOT=$$TRUSTED_BOOT \
-	--build-arg VERSION=${VERSION} .
+	if [ -n "${KUBERNETES_DISTRO}" ]; then \
+		echo "Building standard image with Kubernetes distribution: ${KUBERNETES_DISTRO}"; \
+		docker build --no-cache ${PROGRESS_FLAG} -t ${INIT_IMAGE_NAME} \
+		-f Dockerfile.init \
+		--build-arg BASE_IMAGE=${IMAGE_NAME} \
+		--build-arg TRUSTED_BOOT=$$TRUSTED_BOOT \
+		--build-arg VERSION=${VERSION} \
+		--build-arg KUBERNETES_DISTRO=${KUBERNETES_DISTRO} .; \
```
Itxaka: I think we could reduce this by doing an intermediat EXTRA_FLAGS step in here and just add it to the final command, so we dont repat lines. EXTRA_FLAGS would contain the KUBERNETES_DISTRO and/or KUBERNETES_VERSION or be empty in here

Makefile
jimmykarily: Because I was adding flags and there weren't respected and I thought it might have been the caching. I will try that again, maybe it was something else and it's not needed anymore.

Dockerfile.init
jimmykarily: Good point. We should minimize the duplication in this repository. Building a kairos image in the hadron repo is already a stretch. If we can't keep it to the minimum, we should skip it altogether.

Makefile
Itxaka: ah indeed I just saw this, I think its becuase the image name didnt change between builds so this thinks it didnt change? It should not but no idea.

Dockerfile.init
jimmykarily: Done

Makefile
jimmykarily: I removed it.

Makefile
jimmykarily: done

Makefile
```diff
@@ -11,6 +11,8 @@ KEYS_DIR ?= ${PWD}/tests/assets/keys
 CURRENT_DIR := $(dir $(lastword $(MAKEFILE_LIST)))
 PROGRESS ?= none
 PROGRESS_FLAG = --progress=${PROGRESS}
+KUBERNETES_DISTRO ?=
+KUBERNETES_VERSION ?= latest
```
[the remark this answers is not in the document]
jimmykarily: fixed

Makefile
```diff
@@ -93,11 +96,22 @@ build-kairos:
...
 		TRUSTED_BOOT="false"; \
 	fi; \
-	docker build ${PROGRESS_FLAG} -t ${INIT_IMAGE_NAME} \
-	-f Dockerfile.init \
-	--build-arg BASE_IMAGE=${IMAGE_NAME} \
-	--build-arg TRUSTED_BOOT=$$TRUSTED_BOOT \
-	--build-arg VERSION=${VERSION} .
+	if [ -n "${KUBERNETES_DISTRO}" ]; then \
+		echo "Building standard image with Kubernetes distribution: ${KUBERNETES_DISTRO}"; \
+		docker build --no-cache ${PROGRESS_FLAG} -t ${INIT_IMAGE_NAME} \
```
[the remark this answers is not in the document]
jimmykarily: removed

Makefile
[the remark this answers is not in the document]
jimmykarily: removed

.github/workflows/image_publish.yml
```diff
@@ -322,10 +322,14 @@ jobs:
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@master
+      - name: Download Dockerfile from kairos repository
+        run: |
+          mkdir -p build
+          curl -sSL https://raw.githubusercontent.com/kairos-io/kairos/master/images/Dockerfile -o build/Dockerfile.kairos
```
Itxaka, line 328: noice.

@mauromorales Does the factory action already do this? What I mean, is, can we move the init dockerfile into the factory and still allow overriding it?

jimmykarily: I think we discussed it as some point that the dockerfile belongs to the factory action. I just feel it's going to be a bit "hidden" there and any other consumer who doesn't use github (e.g. using it to build locally) will find it to be a really unexpected location.

mauromorales: Yes, that’s possible. The challenge I see with moving it to the factory is versioning. I think we still want it to remain in the Kairos repo because it should not always download the latest version. If you specify that you want to run Kairos vX.Y.Z, then it should download the corresponding Dockerfile from the Kairos repo, not the latest one.

Of course, your own Dockerfile will always take precedence over those.

.github/workflows/image_publish.yml
```diff
@@ -322,10 +322,16 @@ jobs:
...
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
-          file: ./Dockerfile.init
+          builder: ${{ steps.buildx.outputs.name }}
```
Itxaka, line 332: is this needed? I alwasy though that it would just pick it automatically?

Makefile
```diff
@@ -88,16 +91,28 @@ build-hadron:
...
+	else \
+		echo "Building core image (no Kubernetes distribution)"; \
+		EXTRA_ARGS=""; \
+	fi; \
 	docker build ${PROGRESS_FLAG} -t ${INIT_IMAGE_NAME} \
-	-f Dockerfile.init \
-	--build-arg BASE_IMAGE=${IMAGE_NAME} \
-	--build-arg TRUSTED_BOOT=$$TRUSTED_BOOT \
-	--build-arg VERSION=${VERSION} .
+	  --progress plain \
```
Itxaka: you can set progress to PROGRESS variable as its configurable by the caller so you can either show it or not

Itxaka: in fact the progress is set above so this should be dropped, I guess it was for testing?

Itxaka:
```suggestion
```

Makefile
```diff
@@ -88,16 +91,28 @@ build-hadron:
...
+	@mkdir -p build
+	@curl -sSL https://raw.githubusercontent.com/kairos-io/kairos/master/images/Dockerfile -o build/Dockerfile.kairos || (echo "Error: Failed to fetch Dockerfile from kairos repository" && exit 1)
 	@if [ "${BOOTLOADER}" = "systemd" ]; then \
   		TRUSTED_BOOT="true"; \
 	else \
 		TRUSTED_BOOT="false"; \
 	fi; \
+	if [ -n "${KUBERNETES_DISTRO}" ]; then \
+		echo "Building standard image with Kubernetes distribution: ${KUBERNETES_DISTRO}, version: ${KUBERNETES_VERSION}"; \
+		EXTRA_ARGS="--build-arg KUBERNETES_DISTRO=${KUBERNETES_DISTRO} --build-arg KUBERNETES_VERSION=${KUBERNETES_VERSION}"; \
```
Itxaka: nit: can we maybe call them kubernetes args or something as its specific to that?

Itxaka (approved):
just dropping the extra progress forced flag IMO, rest is goood!

.github/workflows/image_publish.yml
jimmykarily: since we build a builder in the `Set up Docker Buildx` action above, specifying it explicitly makes it more clear that we use that?

Makefile
jimmykarily: yes I was debugging. Removed.

Makefile
jimmykarily: done

Itxaka (approved):

Itxaka:
lets not wait for ci, this looks great!

PR #111, opened 2025-12-03
fix: Improve bashrc and profile-bashrc sourcing

Itxaka:
- Prevent doublesourcing of bashrc by checking BASHRCSOURCED
- Update the path for sourcing the system-wide bashrc file
- Fix loading bashrc on non-login
- Enhance prompt command for better terminal integration

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -989,7 +989,7 @@ RUN cd /sources && \
     --without-bash-malloc \
     --with-installed-readline && make -s -j${JOBS} y.tab.c && make -s -j${JOBS} builtins/libbuiltins.a && make -s -j${JOBS} && \
     mkdir -p /bash/etc/bash && \
-    install -Dm644  /sources/bashrc /bash/etc/bash/bashrc && \
+    install -Dm644  /sources/bashrc /bash/etc/bash.bashrc && \
     install -Dm644  /sources/profile-bashrc.sh /bash/etc/profile.d/00-bashrc.sh && \
     make -s -j${JOBS} DESTDIR=/bash install && make -s -j${JOBS} install # && rm -rf /bash/usr/share/locale
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -484,6 +484,23 @@ ARG GZIP_VERSION=1.12
 RUN cd /sources/downloads && wget -q https://ftp.gnu.org/gnu/gzip/gzip-${GZIP_VERSION}.tar.xz && mv gzip-${GZIP_VERSION}.tar.xz gzip.tar.xz
 
 
+ARG BASH_VERSION=5.3
+# Patch level is the number of patches upstream bash has released for this version https://ftp.gnu.org/gnu/bash/bash-${BASH_VERSION}-patches/
+ARG PATCH_LEVEL=8
+# Get the patches from https://ftp.gnu.org/gnu/bash/bash-${BASH_VERSION}-patches/
+# They are in the format bash$BASH_VERSION_NO_DOT-00$PATCH_LEVEL
+# But the index starts at 1
+
+RUN cd /sources/downloads && wget -q http://mirror.easyname.at/gnu/bash/bash-${BASH_VERSION}.tar.gz && tar -xvf bash-${BASH_VERSION}.tar.gz && mv bash-${BASH_VERSION} bash
+WORKDIR /sources/downloads/bash
+RUN for i in $(seq -w 1 ${PATCH_LEVEL}); do \
+        echo "Applying bash patch bash${BASH_VERSION//./}-00${i}"; \
+        wget -q https://ftp.gnu.org/gnu/bash/bash-${BASH_VERSION}-patches/bash${BASH_VERSION//./}-00${i} -O bash-patch-${i}.patch; \
+        patch -p0 < bash-patch-${i}.patch; \
+    done
+WORKDIR /
+
+
 FROM stage0 AS skeleton
 
 COPY ./setup_rootfs.sh ./setup_rootfs.sh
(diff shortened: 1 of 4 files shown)


mauromorales (approved):
On which system did you test? just wondering if it will work in Ubuntu 🤣

files/bash/bashrc
```diff
@@ -1,13 +1,30 @@
...
-if [ "$PS1" = '\s-\v\$ ' ]; then
-	PS1='\h:\w\$ '
+
+# Prevent doublesourcing
+if [[ -z "${BASHRCSOURCED}" ]] ; then
+  BASHRCSOURCED="Y"
+  # the check is bash's default value
+  [[ "$PS1" = '\s-\v\$ ' ]] && PS1='[\u@\h \W]\$ '
+  case ${TERM} in
+    Eterm*|alacritty*|aterm*|foot*|gnome*|konsole*|kterm*|putty*|rxvt*|tmux*|xterm*)
```
Itxaka, line 14: this is nicer if users log in with one of those.

files/bash/bashrc
```diff
@@ -1,13 +1,30 @@
...
+      PROMPT_COMMAND+=('printf "\033]0;%s@%s:%s\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\~}"')
+      ;;
+    screen*)
+      PROMPT_COMMAND+=('printf "\033_%s@%s:%s\033\\" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\~}"')
+      ;;
+  esac
+fi
+
+if [[ -r /usr/share/bash-completion/bash_completion ]]; then
```
Itxaka, line 23: we currently dont have any completion as we drop them for size but if someone adds them we should load them.

files/bash/profile-bashrc.sh
```diff
@@ -4,5 +4,5 @@ if [ -n "$BASH_VERSION" ] && [ "$BASH" != "/bin/sh" ]; then
 	# and only /etc/profile on login (-l). so, make it do both on login.
 	# this ensures that login-shell bash (e.g. -bash or bash -l) still sources the
 	# system bashrc, which e.g. loads bash-completion
-	. /etc/bash/bashrc
+	. /etc/bash.bashrc
```
Itxaka, line 7: this is what we wanted, install it as the default bash.bashrc so its always loaded

Dockerfile
```diff
@@ -989,7 +989,7 @@ RUN cd /sources && \
     --without-bash-malloc \
     --with-installed-readline && make -s -j${JOBS} y.tab.c && make -s -j${JOBS} builtins/libbuiltins.a && make -s -j${JOBS} && \
     mkdir -p /bash/etc/bash && \
-    install -Dm644  /sources/bashrc /bash/etc/bash/bashrc && \
+    install -Dm644  /sources/bashrc /bash/etc/bash.bashrc && \
```
Itxaka: same here, we wnt this not the dir

files/bash/profile-bashrc.sh
Itxaka: you can check int he source code that this is the default config that it will load: https://github.com/bminor/bash/blob/master/config-top.h#L95

Dockerfile
```diff
@@ -484,6 +484,23 @@ ARG GZIP_VERSION=1.12
 RUN cd /sources/downloads && wget -q https://ftp.gnu.org/gnu/gzip/gzip-${GZIP_VERSION}.tar.xz && mv gzip-${GZIP_VERSION}.tar.xz gzip.tar.xz
 
 
+ARG BASH_VERSION=5.3
+# Patch level is the number of patches upstream bash has released for this version https://ftp.gnu.org/gnu/bash/bash-${BASH_VERSION}-patches/
+ARG PATCH_LEVEL=8
+# Get the patches from https://ftp.gnu.org/gnu/bash/bash-${BASH_VERSION}-patches/
```
Itxaka, line 490: This applies security patches from upstream directly

Dockerfile
```diff
@@ -1050,6 +1038,71 @@ RUN cd /sources && \
...
+RUN rsync -aHAX --keep-dirlinks  /perl/. /
+
+COPY --from=sources-downloader /sources/downloads/bison-${BISON_VERSION}.tar.xz /sources/
+RUN mkdir -p /sources && cd /sources && tar -xvf bison-${BISON_VERSION}.tar.xz && mv bison-${BISON_VERSION} bison && cd bison && mkdir -p /bison && ./configure ${COMMON_CONFIGURE_ARGS} --disable-dependency-tracking --infodir=/usr/share/info --mandir=/usr/share/man --prefix=/usr --disable-static --enable-shared && \
+    make DESTDIR=/bison install && make install
+
+## bash
+FROM readline AS bash
+
+COPY --from=bison /bison /
```
Itxaka, line 1064: Nice to include both bison and flex as deps as they are not required but we can add them for a more full, standard bash. This changes the dockerfile order only, no other changes tot he other targets.

Dockerfile
```diff
@@ -1050,6 +1038,71 @@ RUN cd /sources && \
...
+COPY ./files/bash/profile-bashrc.sh /sources/profile-bashrc.sh
+COPY --from=sources-downloader /sources/downloads/bash /sources/bash
+# If NON_INTERACTIVE_LOGIN_SHELLS is defined, all login shells read the
+# startup files, even if they are not interactive.
+# This makes something like ssh user@host 'command' work as expected, otherwise you would get
+# an error saying that its not a tty
+ENV CFLAGS="${CFLAGS} -DNON_INTERACTIVE_LOGIN_SHELLS -DSSH_SOURCE_BASHRC"
+RUN mkdir -p /bash
+WORKDIR /sources/bash
+RUN CFLAGS="${CFLAGS}" ./configure --quiet ${COMMON_CONFIGURE_ARGS} \
```
Itxaka, line 1077: I needed to pass this directly, otherwise it didnt pick it up

Dockerfile
```diff
@@ -1050,6 +1038,71 @@ RUN cd /sources && \
...
+COPY --from=flex /flex /
+
+COPY ./files/bash/bashrc /sources/bashrc
+COPY ./files/bash/profile-bashrc.sh /sources/profile-bashrc.sh
+COPY --from=sources-downloader /sources/downloads/bash /sources/bash
+# If NON_INTERACTIVE_LOGIN_SHELLS is defined, all login shells read the
+# startup files, even if they are not interactive.
+# This makes something like ssh user@host 'command' work as expected, otherwise you would get
+# an error saying that its not a tty
+ENV CFLAGS="${CFLAGS} -DNON_INTERACTIVE_LOGIN_SHELLS -DSSH_SOURCE_BASHRC"
```
Itxaka, line 1074: This is the real fix, flags so it defines the behaviour when compiling to execute kind of like a login shell when there is no tty asked for

mauromorales (approved):

jimmykarily (approved):

PR #116, opened 2025-12-05
fix: Update multipath-tools and related dependencies

Itxaka:
- Bump MULTIPATH_TOOLS_VERSION from 0.11.3 to 0.13.0
- Bump LVM2_VERSION from 2.03.35 to 2.03.37
- Bump URCU_VERSION from 0.15.2 to 0.15.3
- Adjust build commands for multipath-tools to ensure proper installation

Patch multipathd to avoid a coredump on cleanup.

Seems like while glibc is more careful about freeing a thread when the pthread_join is called, musl is not and calls to an already dereferenced pointer which causes a coredump.

checking the code seems like urcu already calls the pthread_join when call_rcu_data_free is called, so we are doing a double free on the same thread. Fix this by not calling the pthread_join again.

This patch was also sent to upstream to multipath directly as it fixes a real issue which also shows on other musl distros like alpine.

These updates address compatibility issues and improve functionality during cleanup processes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -264,14 +264,14 @@ ENV LIBAIO_VERSION=${LIBAIO_VERSION}
 RUN cd /sources/downloads && wget -q https://pagure.io/libaio/archive/libaio-${LIBAIO_VERSION}/libaio-libaio-${LIBAIO_VERSION}.tar.gz && mv libaio-libaio-${LIBAIO_VERSION}.tar.gz libaio.tar.gz
 
 ## lvm2
-ARG LVM2_VERSION=2.03.35
+ARG LVM2_VERSION=2.03.37
 ENV LVM2_VERSION=${LVM2_VERSION}
 
 RUN cd /sources/downloads && wget -q http://ftp-stud.fht-esslingen.de/pub/Mirrors/sourceware.org/lvm2/releases/LVM2.${LVM2_VERSION}.tgz && mv LVM2.${LVM2_VERSION}.tgz lvm2.tgz
 
 
 ## multipath-tools
-ARG MULTIPATH_TOOLS_VERSION=0.11.3
+ARG MULTIPATH_TOOLS_VERSION=0.13.0
 ENV MULTIPATH_TOOLS_VERSION=${MULTIPATH_TOOLS_VERSION}
 
 RUN cd /sources/downloads && wget -q https://github.com/opensvc/multipath-tools/archive/refs/tags/${MULTIPATH_TOOLS_VERSION}.tar.gz && mv ${MULTIPATH_TOOLS_VERSION}.tar.gz multipath-tools.tar.gz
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1441,13 +1433,13 @@ RUN mv -v gmp-${GMP_VERSION} gdb/gmp
...
-    --target=${TARGET} \
+    --host=${TARGET}  AR=${TARGET}-ar RANLIB=${TARGET}-ranlib NM=${TARGET}-nm CC=${TARGET}-gcc LD=${TARGET}-ld STRIP=${TARGET}-strip \
     --with-sysroot=/ \
     --disable-nls \
     --with-libexpat-prefix=/usr \
     --disable-multilib
-RUN make -s ARCH="${ARCH}" CROSS_COMPILE="${TARGET}-" -j${JOBS}
-RUN make -s ARCH="${ARCH}" CROSS_COMPILE="${TARGET}-" DESTDIR=/gdb install
+RUN make -j${JOBS}
+RUN make -j${JOBS} DESTDIR=/gdb install install-gdbserver
```
Itxaka, line 1442: this now installs a proper gdb+gdbserver for debugging

Dockerfile
```diff
@@ -1982,17 +1923,21 @@ RUN make -s -s && make -s -s install DESTDIR=/iptables
 
 ## libaio for lvm2
 FROM rsync AS libaio
-
+COPY --from=bash /bash /bash
```
Itxaka, line 1926: optional but nice to have

Dockerfile
```diff
@@ -2563,10 +2507,12 @@ WORKDIR /sources
 RUN tar -xf multipath-tools.tar.gz && mv multipath-tools-* multipath-tools
 WORKDIR /sources/multipath-tools
 ENV CC="gcc"
+COPY files/0001-multipathd-Dont-pthread_join-twice.patch /sources/multipath-tools/0001-multipathd-Dont-pthread_join-twice.patch
+RUN patch -p1 </sources/multipath-tools/0001-multipathd-Dont-pthread_join-twice.patch 
 # Set lib to /lib so it works in initramfs as well
-RUN make -s -j${JOBS} WARN_ONLY=1 sysconfdir="/etc" configdir="/etc/multipath/conf.d" LIB=/lib
-RUN make -s -j${JOBS} WARN_ONLY=1 SYSTEMDPATH=/lib LIB=/lib install DESTDIR=/multipath-tools
-RUN make -s -j${JOBS} WARN_ONLY=1 LIB=/lib install
+RUN make -s -j${JOBS} sysconfdir="/etc" configdir="/etc/multipath/conf.d" LIB=/lib
```
Itxaka, line 2513: no need to WARN_ONLY, it builds clean

Dockerfile
```diff
@@ -3267,6 +3237,8 @@ FROM full-image-final AS debug
 
 COPY --from=strace /strace /
 COPY --from=gdb-stage0 /gdb /
+COPY --from=python-build /python /
+RUN --mount=from=gcc-stage0,src=/sysroot/usr/lib,dst=/mnt,ro cp -a /mnt/libstdc++.so* /usr/lib/
```
Itxaka, line 3241: we were missing some deps for gdb in the debug image

mauromorales (approved):

PR #131, opened 2025-12-18
feat: Add Bumpfile for Dockerfile version management

Itxaka:


.github/workflows/autobumper.yml
```diff
@@ -0,0 +1,15 @@
...
+    # minute hour dom month dow (UTC)
+    - cron: '0 9 * * *'
+  # enable manual trigger of version updates
+  workflow_dispatch:
+jobs:
+  bump:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: wader/bump/action@master
```
jimmykarily: Image

76 stars? Are you sure it's the best way to bump versions? It's developed but a single person, I don't see much reviewing happening on the repo.

Itxaka: I asked and already got a comment from the owner, seems to fit our use case perfectly and we can expand it and contribute upstream to it if needed. Its local first, so you can test and run it locally easily to test the changes and see what it will do, seems easy to expand, its been updated recently and they seem to welcome contributions.

Not only that but we dont need to use that action, we can use the bump cli to generate the diff itself and then use something like we do on the packages which creates the PR by itself if we dont want to let a third party creates those (I did set this to try to test manually to see how it creates the PRs and whatnot, not to fully use it, thats why i added the workflow dispatch, to be able to trigger manually)

So ignore the workflow as its only for testing, but the actual command its a bliss to use. Renovate wishes it was this simple

mudler (approved):
love it!

jimmykarily (approved):

Bumpfile
```diff
@@ -1,15 +1,21 @@
...
+# PIPELINE is a filter pipeline: FILTER|FILTER|...
+# A pipeline consist of one or more filters executed in sequence.
+# Usually it starts with a filter that produces versions from some source like a git repository.
+# After that one or more filters can select, transform and sort versions to narrow it down to one version.
+# If a pipeline ends up producing more than one version the first will be used.
 
 # The following section is simple semver checks.
 
 alpine /ALPINE_VERSION=([0-9A-Za-z\.\-]+)/ docker:alpine|^3
-alpine link "Changeset" https://git.alpinelinux.org/aports/log/?h=v$LATEST
```
Itxaka: this only works if we are using the action it seems as it would add this to the commit message

Itxaka:
Im stopping the jobs as they are not affected by this and already passed, lets save cpu cycles

PR #133, opened 2025-12-19
feat: Enhance autobumper configuration for automated version bumps

Itxaka:
- Added environment variables for GitHub authentication and SSH keys
- Implemented cloning and pushing to a forked repository for version updates
- Configured automatic branch creation and PR generation for bumps


.github/workflows/autobumper.yml
```diff
@@ -15,21 +15,61 @@ jobs:
...
+          echo "$GITHUB_PUBLIC_KEY" > $HOME/.ssh/id_rsa.pub
+          chmod 700 ~/.ssh
+          chmod 600 ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa.pub
+
+          git config --global user.name "$GIT_AUTHOR_NAME"
+          git config --global user.email "$GIT_AUTHOR_EMAIL"
+
+          # Ensure fork exists before cloning
+          if ! gh repo view ci-forks/hadron > /dev/null 2>&1; then
```
jimmykarily: can't get make it open a PR from a branch? Do we need the fork?

Itxaka: @mudler wanted to do it like this as it has less privs, so the bot doesnt need any access to the kairos-io repos

.github/workflows/autobumper.yml
```diff
@@ -9,27 +9,118 @@ jobs:
...
-          body: 'Automated version bump'
-          team-reviewers: 'kairos-io/maintainers'
-          base: 'main'
-          add-paths: |
-            Dockerfile
\ No newline at end of file
+          # Check if fork exists, create if not
+          if ! gh repo view ci-forks/hadron --json name > /dev/null 2>&1; then
+            echo "Fork does not exist. Creating fork..."
+            gh repo fork kairos-io/hadron --clone=false --remote=false --org ci-forks
```
mauromorales, line 36: Not sure if this is related

Image

.github/workflows/autobumper.yml
```diff
@@ -9,27 +9,124 @@ jobs:
...
+              fi
+            else
+              echo "No PR found and no changes to commit. Nothing to do."
+            fi
+            exit 0
+          fi
+
+          # Commit and push changes
+          git config user.name "ci-robbot"
+          git config user.email "robot@c3os.io"
```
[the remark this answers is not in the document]
mauromorales: can we change? haha it's a nice observation

.github/workflows/autobumper.yml
Itxaka: yes xDDD I did not check first if ti existed before creating the foirk lmao. But I cant remove them, only @mudler can so....

XDDD

.github/workflows/autobumper.yml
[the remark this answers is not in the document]
Itxaka: yep, but no idea if it would break something. I hat this approach and the one that knows how to set up properly is @mudler

mauromorales (approved):

.github/workflows/autobumper.yml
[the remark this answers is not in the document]
mudler: I can add the email to the user, but It won't break it, it's just configuring git and using a ssh key which is registered in the Github account. This is equivalent of creating a new user, creating a new key, having the private key as secret and controlling a fork where PRs are being opened. If you change the email, the worse thing will happen is that in the commit it won't be shown the user avatar.

mudler (approved):

PR #138, opened 2025-12-21
feat: Add FIPS support in Dockerfile and Makefile

### LINKED ISSUE #3692: feat: Hadron FIPS
This is a tracker for FIPS support for Hadron.

**FIPS Compliance Exploration**  
   - Audit current stack for FIPS 140-2 compatibility (e.g., OpenSSL, crypto modules).  
   - Draft roadmap for FIPS certification:  
     - Phase 1: Isolate crypto modules (e.g., `fips-mode-setup`)  
     - Phase 2: Test against NIST FIPS 140-3 validation suite

Itxaka:
- Introduced FIPS configuration option in Dockerfile
- Added separate build stages for OpenSSL and OpenSSL FIPS
- Updated Makefile to include FIPS variable with default value "no-fips"
- Enhanced OpenSSL installation process to support FIPS mode


jimmykarily (approved):

Itxaka:
```
[root@kairos ~]# cat /proc/sys/crypto/fips_enabled 
1
[root@kairos ~]# cat /etc/os-release 
NAME="Hadron Linux"
PRETTY_NAME="Hadron Linux"
ID=hadron
BUILD_ID=rolling
VERSION_ID="v0.0.1-beta2-13-gb1816bb-dirty"

```

Itxaka:
This will now build a fips version of openssl and use that in the subsequent steps to build against it. Tested with fips=1 in teh cmdline and it boots.

Unfortunately we are still not generating the kernl hmac file but this is almost ready. Next step would be to build the sha512mac and use it to generate a valid .hmac image. Unfortunately dracut searches for the kernel version so we will still end to get the version somehow but the pieces are there

jimmykarily (approved):

Itxaka:
now it has and ships the correct software and hmac to validate the kernel hmac which means that Hadron boots with fips=1 without any modification :D

PR #141, opened 2026-01-02
feat: Refactor Dockerfile to install real kernel headers

Itxaka:
- Replaced the alpine-hack stage with kernel-headers-stage0 for better clarity and structure
- Added logic to handle architecture-specific header installation
- Improved the overall organization of the Dockerfile for easier maintenance

This makes our builds use the actual linux headers for the versions we are using later in the build instead of relaying in the alpine provided ones

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ ENV JOBS=${JOBS}
 ARG MUSSEL_VERSION="95dec40aee2077aa703b7abc7372ba4d34abb889"
 ENV MUSSEL_VERSION=${MUSSEL_VERSION}
 
-RUN apk update && apk add git bash wget bash perl build-base make patch busybox-static curl m4 xz texinfo bison gawk gzip zstd-dev coreutils bzip2 tar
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -482,9 +482,20 @@ RUN <<EOT bash
        make -s -j${JOBS} DESTDIR=/sysroot install ;
 EOT
 
-## This is a hack to avoid to need the kernel headers to compile things like busybox
-FROM alpine:${ALPINE_VERSION} AS alpine-hack
-RUN apk add linux-headers
+FROM make-stage0 AS kernel-headers-stage0
+ARG JOBS
+RUN apk add rsync
```
Itxaka: This is needed as the headers_install uses rsync to copy stuff

mauromorales: maybe add this as comment?

mauromorales (approved):

Dockerfile
Itxaka: I actually moved this to the base container which install all deps, so its easier to track

PR #142, opened 2026-01-04
feat: Optimize build

Itxaka:
- change elfutils by libelf whic is only the needed components to build the kernel
- This results in several components being dropepd as they are no longer needed to build the elfutils
- Enable extra flags in CFLAGS/LDFLAGS for further optimization
- Point AR/NM/RANLIB to the gcc wrappers instead of the binutils directly so they understand the flto flags
- Disable uneeded and weak cyphers on openssl
- disable nls is several components that were missing the flag
- disable uneeded flags and features in iptables
- disable uneeded flags and features in tpm2 libs
- disable further unused components in systemd
- remove bash/zsh/fish completions
- remove useless keymaps
- remove any leftover locales that there may be left
- remove any pkconfig or middle built stuff if leftover

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -150,10 +150,6 @@ RUN wget -q https://github.com/westes/flex/releases/download/v${FLEX_VERSION}/fl
 ARG BISON_VERSION=3.8.2
 RUN wget -q https://ftpmirror.gnu.org/bison/bison-${BISON_VERSION}.tar.xz -O bison.tar.xz
 
-## argp-standalone
-ARG ARGP_STANDALONE_VERSION=1.3
-RUN wget -q http://www.lysator.liu.se/~nisse/misc/argp-standalone-${ARGP_STANDALONE_VERSION}.tar.gz -O argp-standalone.tar.gz
-
 ## autoconf
 ARG AUTOCONF_VERSION=2.72
 RUN wget -q https://ftpmirror.gnu.org/autoconf/autoconf-${AUTOCONF_VERSION}.tar.xz -O autoconf.tar.xz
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -172,14 +168,8 @@ RUN wget -q https://github.com/pullmoll/musl-fts/archive/v${FTS_VERSION}.tar.gz
...
-## musl-obstack
-ARG MUSL_OBSTACK_VERSION=1.2.3
-RUN wget -q https://github.com/void-linux/musl-obstack/archive/v${MUSL_OBSTACK_VERSION}.tar.gz -O musl-obstack.tar.gz
-
-## elfutils
-ARG ELFUTILS_VERSION=0.194
-RUN wget -q https://sourceware.org/elfutils/ftp/${ELFUTILS_VERSION}/elfutils-${ELFUTILS_VERSION}.tar.bz2 -O elfutils.tar.bz2
-RUN mkdir -p elfutils-patches && wget -q https://gitlab.alpinelinux.org/alpine/aports/-/raw/master/main/elfutils/musl-macros.patch -O elfutils-patches/musl-macros.patch
+ARG LIBELF_VERSION=0.193
+RUN wget -q https://github.com/arachsys/libelf/archive/refs/tags/v0.193.tar.gz -O libelf.tar.gz
```
mudler, line 172: this is a nice hack, just wondering if we can relay on this dependency rather than classic elfutils. However, in any case it shouldn't be a big issue either way.

mudler (approved):

Dockerfile
Itxaka: If not, we could probably not build anything as it seems like we only need the geld.h files so, basically the headers obly.

PR #145, opened 2026-01-07
Create DEV.md

Itxaka:
some comments on helpful things

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -4,13 +4,13 @@ Some comments about dev here.
 # Size reduction in OS components
 
 - Try to compile everything with `-Os` for make or in the case of meson builds `--buildtype=minsize`
-- We strip at the end, but fi you build with meson, there is a swtich to do it automatically which works nicely `-Dstrip=true` and valid for all meson builds
-- If you have a single dep only used by a single program, build it static. If its not used by any other thing int he final system, its much nicer to incrust whatever is needed in the final app rather than have the full shared lib which is only used by a single program.
+- We strip at the end, but if you build with meson, there is a switch to do it automatically which works nicely `-Dstrip=true` and valid for all meson builds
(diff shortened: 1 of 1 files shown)


DEV.md
```diff
@@ -0,0 +1,17 @@
+Some comments about dev here.
+
+
+# Size reduction in OS components
+
+- Try to compile everything with `-Os` for make or in the case of meson builds `--buildtype=minsize`
+- We strip at the end, but fi you build with meson, there is a swtich to do it automatically which works nicely `-Dstrip=true` and valid for all meson builds
```
jimmykarily:
```suggestion
- We strip at the end, but if you build with meson, there is a switch to do it automatically which works nicely `-Dstrip=true` and valid for all meson builds
```

DEV.md
```diff
@@ -0,0 +1,17 @@
...
+- If you have a single dep only used by a single program, build it static. If its not used by any other thing int he final system, its much nicer to incrust whatever is needed in the final app rather than have the full shared lib which is only used by a single program.
```
jimmykarily:
```suggestion
- If you have a single dep only used by a single program, build it static. If its not used by any other thing in the final system, its much nicer to encrust whatever is needed in the final app rather than have the full shared lib which is only used by a single program.
```

DEV.md
```diff
@@ -0,0 +1,17 @@
...
+- Try to reduce build options to what we need, for example TPM2 lib builds a lot of libraries by default but we dont need but a small subset to support luks encryption via TPM, so try to check the configure options and be selective
+
+# golang size reduction
+
+- Comrpess everything with upx `--best --lzma` for final releases
```
jimmykarily:
```suggestion
- Compress everything with upx `--best --lzma` for final releases
```

jimmykarily (approved):

PR #149, opened 2026-01-19
chore: Update systemd version handling in Bumpfile

Itxaka:
- Changed systemd version specification from fixed `257` to use `semver:*` for flexibility
- This allows for automatic updates to newer systemd versions as they become compatible


Bumpfile
```diff
@@ -33,9 +33,7 @@ attr /ATTR_VERSION=([0-9A-Za-z\.\-]+)/ fetch:https://download.savannah.nongnu.or
 
 gawk /GAWK_VERSION=([0-9A-Za-z\.\-]+)/ fetch:https://ftpmirror.gnu.org/gawk/|/gawk-([0-9\.]+)\.tar\.xz/|semver:*
 
-# Stick to systemd 257 for now as we dont have patches for later versions to build with musl and 259+musl is not ready yet
```
jimmykarily, line 36: no longer true then?

jimmykarily (approved):

Bumpfile
Itxaka: yep, we now have a small patch for musl that makes systemd build clean

PR #169, opened 2026-02-05
Update coreutils

Itxaka:
https://lists.gnu.org/archive/html/coreutils-announce/2026-02/msg00000.html

Notable changes include:
- Options in man pages link directly into the full web docs
- timeout(1) now kills the command for all terminating signals
- paste(1) is now multi-byte character aware
- cp(1) fixes an unlikely infinite loop introduced in v9.9
- The multi-call binary is 3.2% smaller


mudler:
Superseded by https://github.com/kairos-io/hadron/pull/186

PR #215, opened 2026-02-26
fix: Update core-system.yaml and Dockerfile.nvidia

Itxaka:
- Modified command in core-system.yaml to filter out release candidates
- Added sorting and semver constraints to ensure proper version handling
- Updated comments in Dockerfile.nvidia for clarity on glibc library handling
- Included Dockerfile.nvidia in examples.yml for better visibility

### DIFF
--- a/updatecli.d/core-system.yaml
+++ b/updatecli.d/core-system.yaml
@@ -23,10 +23,14 @@ sources:
     spec:
       command: |
         git ls-remote --refs --tags https://github.com/systemd/systemd.git | grep -v '\^{}' \
-        | awk -F/ '{print $NF}' \
+        | awk -F/ '{print $NF}' | grep -v rc | grep -v RC \
         | sort -V | tail -n 1
     transformers:
       - trimprefix: v
+      - sort:
+          method: semver
+      - semver:
+          constraint: '*'
 
   util_linux:
     kind: githubrelease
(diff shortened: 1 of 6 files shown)


updatecli.d/core-system.yaml
```diff
@@ -23,10 +23,14 @@ sources:
     spec:
       command: |
         git ls-remote --refs --tags https://github.com/systemd/systemd.git | grep -v '\^{}' \
-        | awk -F/ '{print $NF}' \
+        | awk -F/ '{print $NF}' | grep -v rc | grep -v RC \
```
mauromorales, line 26: I think this should work, not sure though :P 

```suggestion
        | awk -F/ '{print $NF}' | grep -vi 'rc' \
```

mauromorales: ah no but that might mean rC or Rc ... ok forget it ;P

Itxaka: I mean, this is just a workaround for now, ideally it should have a condition on each source to make sure that we validate no rc/beta/alpha/whatever

But thats a lot of work to do :D

mauromorales (approved):

Itxaka:
only affects examples building and updates PRs so merging this directly to avoid wating cpu time

PR #252, opened 2026-03-12
feat: enable Intel TDX host support in default kernel

mudler:


Itxaka:
default or cloud or both?

Itxaka (approved):
for me this also enabled CONFIG_KVM_INTEL_TDX

jimmykarily:
For AMD, chatgpt suggests these too:


```
CONFIG_AMD_MEM_ENCRYPT=y
CONFIG_AMD_SEV=y
CONFIG_AMD_SEV_SNP=y
CONFIG_KVM_AMD_SEV=y
```

we don't seem to have them all.

jimmykarily:
@copilot check the comments in this PR and open another PR addressing those if you think it makes sense.

jimmykarily (approved):

Itxaka:
btw I approved this but im not sure we should be doing manual edits of the config files. There is a loot of tooling around this to prevent mistakes and bring dependant stuff or enable things that the entries are dependant on, so going in blind and just editing the file is bad IMHO. either use the scripts under the linux src which enable stuff or use make menuconfig to see the entries and enable dependants.

mudler:
> btw I approved this but im not sure we should be doing manual edits of the config files. There is a loot of tooling around this to prevent mistakes and bring dependant stuff or enable things that the entries are dependant on, so going in blind and just editing the file is bad IMHO. either use the scripts under the linux src which enable stuff or use make menuconfig to see the entries and enable dependants.

to be fair, when maintaining kernels I've always leaned to another approach, go with the minimum config you need ( no need to select from the interface ) and then run make olddefconfig so you get best of both worlds: the defaults for *that* current kernel would get populated. This is because If you enable all the options that the menuconfig gives you, that might change between versions, and you might not need anymore flags that you have enabled additionally. Minimum worked best for me for long-term maintenance.

Itxaka:
> > btw I approved this but im not sure we should be doing manual edits of the config files. There is a loot of tooling around this to prevent mistakes and bring dependant stuff or enable things that the entries are dependant on, so going in blind and just editing the file is bad IMHO. either use the scripts under the linux src which enable stuff or use make menuconfig to see the entries and enable dependants.
> 
> to be fair, when maintaining kernels I've always leaned to another approach, go with the minimum config you need ( no need to select from the interface ) and then run make olddefconfig so you get best of both worlds: the defaults for _that_ current kernel would get populated. This is because If you enable all the options that the menuconfig gives you, that might change between versions, and you might not need anymore flags that you have enabled additionally. Minimum worked best for me for long-term maintenance.
> > btw I approved this but im not sure we should be doing manual edits of the config files. There is a loot of tooling around this to prevent mistakes and bring dependant stuff or enable things that the entries are dependant on, so going in blind and just editing the file is bad IMHO. either use the scripts under the linux src which enable stuff or use make menuconfig to see the entries and enable dependants.
> 
> to be fair, when maintaining kernels I've always leaned to another approach, go with the minimum config you need ( no need to select from the interface ) and then run make olddefconfig so you get best of both worlds: the defaults for _that_ current kernel would get populated. This is because If you enable all the options that the menuconfig gives you, that might change between versions, and you might not need anymore flags that you have enabled additionally. Minimum worked best for me for long-term maintenance.

well we run olddefconfig as part of the build, but I know there are some bits and bops that are not enabled, no matter how many times you put KEY=y if others are not enabled before, so manually editing this could lead to us thinking we have support for X but we dont. 

For example, for this specific case the `CONFIG_INTEL_TDX_HOST` needs `CPU_SUP_INTEL [=y] && X86_64 [=y] && KVM_INTEL [=m] && X86_X2APIC [=y] && CONTIG_ALLOC [=y] && X86_MCE [=y]`

In this case those are already enabled, but if they were not, I dont think they will be enabled even if you run an oldconfig, they just wont and in fact probable the `CONFIG_INTEL_TDX_HOST` will be disabled. Same with modules, some do not accept being modules and are either y/n so a wrong value + olddefconfig will turn those to their default.


Just saying, this is more complex than an AI just adding KEY=VALUE to a file, I dont feel secure enough to let it do its own thing without human supervision.

mudler:
> Just saying, this is more complex than an AI just adding KEY=VALUE to a file, I dont feel secure enough to let it do its own thing without human supervision.

But this is not running AI without supervision. PRs are made for being reviewed, and we actually were just checking what it would do, as we were looking at the documentation.

By the way - this PR was run manually by editing the file from Github :) Probably you were refering to https://github.com/kairos-io/hadron/pull/256

PR #275, opened 2026-03-23
fix: Update arm64 kernel configuration

Itxaka:
- Updated kernel version from 6.18.4 to 6.19.9
- Changed GCC version from 15.2.1 to 15.1.0
- Adjusted assembler and linker versions for compatibility
- Enabled CONFIG_KEXEC_HANDOVER and related options for improved boot handling
- Added new configuration options for enhanced hardware support (mainly Tegra devices)


files/kernel/default-arm64.config
```diff
@@ -3552,6 +3582,8 @@ CONFIG_VCAP=y
 # CONFIG_NET_VENDOR_MICROSEMI is not set
 CONFIG_NET_VENDOR_MICROSOFT=y
 CONFIG_MICROSOFT_MANA=m
+CONFIG_NET_VENDOR_MUCSE=y
```
mudler: should be probably `m`?

files/kernel/default-arm64.config
```diff
@@ -9954,6 +10048,8 @@ CONFIG_ARM64_PLATFORM_DEVICES=y
 # CONFIG_EC_HUAWEI_GAOKUN is not set
 # CONFIG_EC_LENOVO_YOGA_C630 is not set
 # CONFIG_EC_LENOVO_THINKPAD_T14S is not set
+CONFIG_BCM_VIDEOCORE=m
```
mudler, line 10050: good catch, this was missed from Rpi?

mudler (approved):

files/kernel/default-arm64.config
Itxaka: this was enabled by default by the updated 6.19.9 config. I think we had a very old default in there (very old lmao, 6.18 xD)

files/kernel/default-arm64.config
Itxaka: no idea where this came from, I updated only the tegra and the rest is probably the defconfig?

Its either on or none, no module for this.

Itxaka:
I canceled the amd64 path for now as its not needed, this only touches arm

PR #276, opened 2026-03-23
feat: Bump MUSL version to 1.2.6

Itxaka:
- Updated `MUSL_VERSION` from 1.2.5 to 1.2.6 to incorporate the latest features and improvements.
- This change ensures compatibility with newer software dependencies and enhances overall performance.
- This also allows us to drop the single patch we used to maintain musl compatibility with LIBC


patches/0001-musl-stdio-skipempty-iovec-when-buffering-is-disabled.patch
```diff
@@ -1,30 +0,0 @@
-From fde29c04adbab9d5b081bf6717b5458188647f1c Mon Sep 17 00:00:00 2001
```
mudler, line 1: feels so good to drop patches !

mudler (approved):

PR #307, opened 2026-04-11
feat: add RISC-V (riscv64) architecture support

mauromorales:
- Add PR_riscv64.yml workflow using RISE RISC-V runners for native builds
- Update Dockerfile with RISC-V cross-compilation, kernel builds, GRUB EFI
- Update Makefile to support riscv64 arch and QEMU emulation
- Add cloud and default kernel configs for RISC-V

Made-with: Cursor


Dockerfile
```diff
@@ -41,8 +41,12 @@ ENV MUSSEL_VERSION=${MUSSEL_VERSION}
 # Validate that the arches are correct
 RUN if [ "${ARCH}" = "x86-64" ] && [ "${BUILD_ARCH}" != "x86_64" ]; then echo "For ARCH x86-64, BUILD_ARCH must be x86_64"; exit 1; fi
 RUN if [ "${ARCH}" = "aarch64" ] && [ "${BUILD_ARCH}" != "aarch64" ]; then echo "For ARCH aarch64, BUILD_ARCH must be aarch64"; exit 1; fi
+RUN if [ "${ARCH}" = "riscv64" ] && [ "${BUILD_ARCH}" != "riscv64" ]; then echo "For ARCH riscv64, BUILD_ARCH must be riscv64"; exit 1; fi
 
 RUN git clone https://github.com/firasuke/mussel.git && cd mussel && git checkout ${MUSSEL_VERSION} -b build
+# Fix mussel RISC-V configuration: remove unsupported --with-cpu option for RISC-V
+# GCC for RISC-V doesn't support --with-cpu, only --with-arch, --with-tune, and --with-abi
+RUN cd mussel && sed -i 's/--with-cpu=sifive-u74 --with-arch=rv64gc --with-tune=sifive-7-series/--with-arch=rv64gc --with-tune=rocket/g' mussel
```
Itxaka: dont get this? Why not subs `--with-cpu=sifive-u74` with `--with-mcpu=sifive-u74` and leave the others as is?

Dockerfile
```diff
@@ -643,17 +647,20 @@ COPY --from=sources-downloader /sources/downloads/busybox.tar.bz2 /sources/
 RUN cd /sources && tar -xf busybox.tar.bz2 && \
     mv busybox-* busybox && cd busybox && \
     make -s distclean && \
-    make -s ARCH="${ARCH}" defconfig && \
+    BUSYBOX_ARCH="x86_64" && \
```
Itxaka: I had no issues with the arch itself, its just that you need to disable CONFIG_SHA1_HWACCEL everywhere that its not x86, the rest I did not need ot change for it to work.

Dockerfile
mauromorales: @Itxaka for what I understand there's no `--with-mcpu` at configuration option, but I know too little about this so feel free to explain further.

I changed so it's more generic and not sifive oriented, not sure how common sifive is but I thought it might still be worth making it more generic, maybe we can ask some feedback later from someone from the RISC-V community

Itxaka: upstream mussle says that just remove the --with-cpu entry and leave the4 reset as is :D

Itxaka: mainly becuase we dont know what the generic risc is, I would leave as it is for now to not touch/broke anything for now./ In any case, this is the generic riscv cross compiler than then we drop so maybe its not a big issue?

mauromorales: ok, running locally to be sure it works

Itxaka: ah it seems that you can drop this then override the cpu later by using `CFLAGS` to pass `-mcpu=sifive-u74` or others.

.github/workflows/PR_riscv64.yml
```diff
@@ -0,0 +1,607 @@
...
+          builder: ${{ steps.buildx.outputs.name }}
+          context: ./
+          file: ./Dockerfile
+          platforms: linux/riscv64
+          labels: ${{ steps.meta.outputs.labels }}
+          target: stage0
+          cache-from: |
+            type=registry,ref=quay.io/kairos/hadron-build-cache:stage0-riscv64,mode=max
+          cache-to: |
+            type=registry,ref=quay.io/kairos/hadron-build-cache:stage0-riscv64,mode=max,ignore-error=true,compression=zstd,compression-level=3,oci-mediatypes=true,image-manifest=true
```
Itxaka: cache to from PRs means that we need to expose the user/pass on prs AND external PRS will always fail AND the cache will always be invalidated due PR pushing a cache that its not valid. Cache is consumed on PR and never pushed, only on main its pushed :D

Dockerfile
```diff
@@ -41,8 +41,11 @@ ENV MUSSEL_VERSION=${MUSSEL_VERSION}
 # Validate that the arches are correct
 RUN if [ "${ARCH}" = "x86-64" ] && [ "${BUILD_ARCH}" != "x86_64" ]; then echo "For ARCH x86-64, BUILD_ARCH must be x86_64"; exit 1; fi
 RUN if [ "${ARCH}" = "aarch64" ] && [ "${BUILD_ARCH}" != "aarch64" ]; then echo "For ARCH aarch64, BUILD_ARCH must be aarch64"; exit 1; fi
+RUN if [ "${ARCH}" = "riscv64" ] && [ "${BUILD_ARCH}" != "riscv64" ]; then echo "For ARCH riscv64, BUILD_ARCH must be riscv64"; exit 1; fi
 
 RUN git clone https://github.com/firasuke/mussel.git && cd mussel && git checkout ${MUSSEL_VERSION} -b build
+# Fix mussel RISC-V configuration: remove unsupported --with-cpu option
+RUN cd mussel && sed -i 's/--with-cpu=sifive-u74 //g' mussel
```
Itxaka: this is fixed on https://github.com/firasuke/mussel/commit/687d2f5e4d679487209cd9b4bd75091a20cae357 we could try to bump it to that commit to see if it works out of the box now.

Dockerfile
```diff
@@ -2337,13 +2361,13 @@ ARG CFLAGS="${CFLAGS//-flto=auto/}"
 ARG LDFLAGS="${LDFLAGS//-flto=auto/}"
 WORKDIR /sources/grub
 RUN mkdir -p /grub-bios
-# Protect against building grub-bios on aarch64 host which is not supported
-RUN if [ "${ARCH}" != "aarch64" ]; then ./configure ${COMMON_CONFIGURE_ARGS} --with-platform=pc --disable-werror;fi
+# Protect against building grub-bios on aarch64/riscv64 host which is not supported
+RUN if [ "${ARCH}" != "aarch64" ] && [ "${ARCH}" != "riscv64" ]; then ./configure ${COMMON_CONFIGURE_ARGS} --with-platform=pc --disable-werror;fi
```
Itxaka: isnt this simpler for now to check if arch == x86 xD?

Dockerfile
```diff
@@ -2337,13 +2361,13 @@ ARG CFLAGS="${CFLAGS//-flto=auto/}"
...
 # Reconfigure gnulib shipped with grub to avoid build issues
 # This comes because on grub 2.14 these files are shipped pre-generated and they were built on a glibc system
 # which causes issues when building on musl systems as it expects the bsd-compat-headers to be available
 # which is not the case here. So we force regenerating these files with our musl toolchain so it can find there is no cdefs
-RUN if [ "${ARCH}" != "aarch64" ]; then make -s -j${JOBS} -l${MAX_LOAD} -C grub-core/lib/gnulib;fi
+RUN if [ "${ARCH}" != "aarch64" ] && [ "${ARCH}" != "riscv64" ]; then make -s -j${JOBS} -l${MAX_LOAD} -C grub-core/lib/gnulib;fi
```
Itxaka: ditto

mauromorales:
fyi @luhenry, and any feedback is also welcome

PR #313, opened 2026-04-15
Build raspberry pi dtbs

jimmykarily:
to make sdcard work won Raspberry pi 4

Relates to: https://github.com/kairos-io/kairos/issues/4029

### DIFF
--- a/.github/workflows/PR_arm64.yml
+++ b/.github/workflows/PR_arm64.yml
@@ -107,7 +107,7 @@ jobs:
       - container  # Take advantage of cached layers from the container build
     strategy:
       matrix:
-        kernel: [ "cloud" ]
+        kernel: [ "cloud", "default" ]
     permissions:
       packages: write
       contents: read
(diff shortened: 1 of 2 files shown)


jimmykarily:
I want CI to build an image quickly (on real arm) so that I can test this fix.

Itxaka:
not sure how this can be the problem. Dont we ship the dbts already when creating the image? Wouldnt missing dbts also affect any other distros like ubuntu ?

jimmykarily:
> not sure how this can be the problem. Dont we ship the dbts already when creating the image? Wouldnt missing dbts also affect any other distros like ubuntu ?

I assume other distros ship dtbs as part of some package which we include (e.g. https://packages.ubuntu.com/search?searchon=contents&keywords=bcm2711-rpi-4-b.dtb&arch=arm64&suite=jammy). 

I can't find anywhere in Hadron where we include the dtbs. Got a link?

Itxaka:
> > not sure how this can be the problem. Dont we ship the dbts already when creating the image? Wouldnt missing dbts also affect any other distros like ubuntu ?
> 
> I assume other distros ship dtbs as part of some package which we include (e.g. [packages.ubuntu.com/search?searchon=contents&keywords=bcm2711-rpi-4-b.dtb&arch=arm64&suite=jammy](https://packages.ubuntu.com/search?searchon=contents&keywords=bcm2711-rpi-4-b.dtb&arch=arm64&suite=jammy)).
> 
> I can't find anywhere in Hadron where we include the dtbs. Got a link?

We dont include the dtbs anywhere but in Aurora, so anything building for rpi gets those dtbs + uboot in there form a single source: https://github.com/kairos-io/AuroraBoot/blob/main/Dockerfile#L104 and the package just brings the contents of https://github.com/raspberrypi/firmware

So really it should not be anything to do with building the dtbs, maybe the modules are not in initramfs or the kernel does not load the dtb? (Im not even sure the dbt are needed when using uboot, I though it had them internally or something...

jimmykarily:
It's true that AuroraBoot handles this. Still the dtbs shipping with uboot don't seem to enable the mmc device. I don't know what's wrong at this point. On an dracut emergency shell, `lsblk` doesn't return any device.

PR #18, opened 2025-09-25
Install glibc Kernel and set some basic things for a workable artifact

Itxaka:
- Adds tiny kernel from luet packages
- Adjust symlink creation in setup_rootfs.sh for relative paths
- Add download commands for new dependencies: dbus, expat, seccomp, strace, and kbd in Dockerfile
- Change download mirrors for various packages to improve reliability
- Update kernel installation command to use linux-tiny
- Build systemd with support for dbus and seccomp, otherwise half of the services wont start
- Create 2 different targets:
  - stage2-test will generate an image "ready" to be consumed by auroraboot and build and iso that boots. It has immucore and a custom init that mounts the livecd, agent and so on, but lacks any kind of kairos configs or anything. Its basically an image for testing in qemu easily and to get a running environment
  - stage2 will generate an image with the software installed and the kernel but no kairos parts, no initramfs. This should be the final image pushed, like other distros do. This is the image that kairos-init would ingest and transform into a kairos image so we need only the basics for it.

PR #336, opened 2026-04-23
Enable Confidential Computing needed options in kernel compilation

jimmykarily:
This is a test PR to quickly get ttl.sh artifacts built. I will manually QA on AWS before we decide how we want to maintain these options (See [here for more](#18 2.187 /usr/lib/dracut/modules.d/10systemd/module-setup.sh: line 137: ldd: command not found)).


jimmykarily:
Some links Claude collected that worth looking into:

```
The most relevant — shared GHCB page during early memory acceptance (April 2025):
  https://lkml.org/lkml/2025/4/17/1283

  RFC: early boot SNP panic (February 2026):
  https://lkml.org/lkml/2026/2/25/2136
  https://lkml.org/lkml/2026/2/26/1126 (Tom Lendacky's response)

  CVE-2026-31591 (host-side KVM SNP race, fixed in 6.19.14):
  https://windowsforum.com/threads/cve-2026-31591-linux-kvm-amd-sev-snp-vcpu-locking-race-can-crash-hosts.415220/

  The April 2025 LKML thread is the one most likely relevant to our exact symptom (silent hang during EFI boot with unaccepted memory + SNP). Worth
   checking if that fix landed in 6.19.12 or only later.
```

It suggests our problem might be the a bug in the specific kernel we use in Hadron `6.19.12`  . Bumping to latest might help. Needs checking and testing.

PR #342, opened 2026-04-25
feat: Override root shell to /bin/bash in Docker configuration

Itxaka:
- Added configuration to set the root user's shell to /bin/bash
- Created new file `00-root.conf` for user and group definitions
- This change ensures compatibility with systemd's default settings

### DIFF
--- a/files/systemd/00-root.conf
+++ b/files/systemd/00-root.conf
@@ -1,0 +1,2 @@
+u root 0:0 "Super User" /root /bin/bash
+g root 0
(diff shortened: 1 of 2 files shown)


jimmykarily (approved):

files/systemd/00-root.conf
```diff
@@ -0,0 +1,2 @@
+u root 0:0 "Super User" /root /bin/bash
+g root 0
```
jimmykarily, line 2: Is this line redundant with the second zero above?

jimmykarily (approved):

files/systemd/00-root.conf
Itxaka: no, this creates the group for root. Not really needed per se as its the same as in the default config, but for completition I added it so this whole file is the one that configures the root user+group

Itxaka:
Tested and root user is create with the proper shell

PR #364, opened 2026-05-11
fix: use native build flags in example Dockerfiles

mauromorales:
## Summary

- Replace `${COMMON_CONFIGURE_ARGS}` with explicit `--quiet --prefix=/usr` in example Dockerfiles
- Remove `--build=${BUILD}` references that were causing cross-compilation detection

The `COMMON_CONFIGURE_ARGS` variable sets mismatched `--host` and `--build` triplets (`x86_64-hadron-linux-musl` vs `x86_64-pc-linux-musl`), causing autoconf to treat builds as cross-compilation and refuse to run test programs with:

```
configure: error: cannot run test program while cross compiling
```

Since these are examples that only need to build on a single architecture, native build flags are sufficient.

Fixes the build failure in https://github.com/kairos-io/hadron/actions/runs/25685218794/job/75407036479

## Test plan

- [ ] Verify CI builds pass for all example Dockerfiles (doom, fwupd, git, gpg)

Made with [Cursor](https://cursor.com)


Itxaka (commented):
Im not sure this IS correct. The flag IS set by the tool chain itself so if it's hard coded to x86 then the problem is in there's on our setting of it. Having it as a flag means that when changing the flags in the future, if we ever did, it would trickle down

Itxaka:
closing it for now, lets revisit because we did some changes to flagsz and whatnot so this may be fixed?

PR #370, opened 2026-05-12
hadron-splash

Itxaka:
Adds a small binary that runs an animated hadron logo and shows the version
output


Dockerfile
```diff
@@ -3223,6 +3231,8 @@ RUN rsync -aHAX --keep-dirlinks  /kbd/. /skeleton
 COPY --from=openssl /openssl /openssl
 RUN rsync -aHAX --keep-dirlinks  /openssl/. /skeleton/
 
+COPY --from=hadron-splash /hadron-splash/hadron /skeleton/bin/hadron
```
mudler: I'd suggest to call it `hadron-splash`, I would reserve `hadron` for later

mauromorales (approved):
awesome 😎

PR #380, opened 2026-05-19
Automatic version bumps for: 

ci-robbot:


Itxaka:
@copilot i think this new versión requieres us to deop the útil Linux patch that we had or rrwork It. Please test and create a PR to deop It as part of this versión bump

PR #388, opened 2026-05-26
feat: open-vm-tools

Itxaka:
Plus some improvements to main toolchain image

### DIFF
--- a/examples/add-packages/Dockerfile.open-vm-tools
+++ b/examples/add-packages/Dockerfile.open-vm-tools
@@ -1,0 +1,178 @@
+ARG HADRON_VERSION=main
+ARG JOBS=4
+
+FROM ghcr.io/kairos-io/hadron-toolchain:${HADRON_VERSION} AS downloader
+RUN pip3 install meson ninja
+RUN mkdir -p /output
+WORKDIR /sources
+
+ARG OPENVM_TOOLS_MAJOR=13.1.0
+ARG OPENVM_TOOLS_VERSION=13.1.0-25218885
+RUN curl -L https://github.com/vmware/open-vm-tools/releases/download/stable-${OPENVM_TOOLS_MAJOR}/open-vm-tools-${OPENVM_TOOLS_VERSION}.tar.gz -o openvm-tools.tar.gz
+
+ARG MSPACK_VERSION=1.11
+RUN curl -L https://github.com/kyz/libmspack/archive/refs/tags/v${MSPACK_VERSION}.tar.gz -o mspack.tar.gz
+
+ARG GLIB_VERSION=2.86.2
+RUN curl -L https://download.gnome.org/sources/glib/2.86/glib-${GLIB_VERSION}.tar.xz -o glib.tar.xz
+
+ARG PCRE2_VERSION=10.47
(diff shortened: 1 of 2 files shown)


examples/add-packages/Dockerfile.open-vm-tools
```diff
@@ -0,0 +1,178 @@
...
+FROM downloader AS libtirpc
+ARG JOBS
+COPY --from=bsd-compat-headers /output/ /
+WORKDIR /sources/libtirpc
+
+RUN ./configure ${COMMON_CONFIGURE_ARGS} --disable-gssapi
+RUN make -s -j${JOBS}
+RUN make -s -j${JOBS} install DESTDIR=/output
+
+# TRhis package is a workaround for rpcgen
```
jimmykarily:
```suggestion
# This package is a workaround for rpcgen
```

examples/add-packages/Dockerfile.open-vm-tools
```diff
@@ -0,0 +1,178 @@
...
+# Binary usually comes fomr libgc but in our case libtirpc should substitute it, so we need the rpcgen binary
+# This only compiles that part.
+FROM downloader AS rpcsvc-proto
+ARG JOBS
+RUN curl -L https://github.com/thkukuk/rpcsvc-proto/releases/download/v1.4.4/rpcsvc-proto-1.4.4.tar.xz -o rpcsvc-proto.tar.xz
```
jimmykarily, line 90: Do we generally hardcode versions like this? Shouldn't this be an ARG or something?

examples/add-packages/Dockerfile.open-vm-tools
```diff
@@ -0,0 +1,178 @@
...
+FROM downloader AS openvmtools
+ARG JOBS
+COPY --from=pcre2 /output/ /
+COPY --from=libffi /output/ /
+COPY --from=glib2 /output/ /
+COPY --from=libmspack /output/ /
+COPY --from=libtirpc /output/ /
+COPY --from=rpcsvc-proto /output/ /
+# Its patching time!
+# we should really get with alpine and postmarketOS and try to push some of this patches upstream because this is a shame
```
jimmykarily, line 106: Where did we get this list from? Alpine or postmarketOS maybe? Might worth mentioning here so we can compare in the future is we need to update this list of patches.

jimmykarily (approved):

examples/add-packages/Dockerfile.open-vm-tools
Itxaka: This will be removed when we move this into cloud so Im leaving it as is to not trigger a new run

examples/add-packages/Dockerfile.open-vm-tools
Itxaka: from alpine indeed, but again when moved intothe main dockerfile this is moot as this is part of the apportsd package so we will have them locally :D

examples/add-packages/Dockerfile.open-vm-tools
Itxaka: yeah I was bored of doing that just for a temp file :D

Itxaka:
will wait for https://github.com/kairos-io/hadron/pull/389 to get in as to not break anything

PR #390, opened 2026-05-28
feat: Bundle vmtools and qemu-ga with cloud images

Itxaka:
- Introduced separate download stages for PCRE2, GLib, QEMU, and MSPack in the Dockerfile to improve build organization and caching.
- Each stage fetches the respective source tarballs, allowing for parallel downloads and better management of dependencies.
- Updated the final image to include these components, ensuring that the necessary libraries are available for the build process.

This provides the cloud tools in the cloud images as expected, like wiht the kernel configs, so everything works out of the box when running into a cloud provider like kvm, vsphere and whatnot

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3392,8 +3392,8 @@ WORKDIR /output
 
 # Map build arches to the stage consumed by cloud-tools
 FROM open-vm-tools-build AS open-vm-tools-amd64
-FROM open-vm-tools-build AS open-vm-tools-aarch64
-FROM open-vm-tools-${BUILDARCH} AS open-vm-tools
+FROM open-vm-tools-build AS open-vm-tools-arm64
+FROM open-vm-tools-${TARGETARCH} AS open-vm-tools
 
 
 FROM rsync AS cloud-tools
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -3565,6 +3746,8 @@ RUN if [ "${ARCH}" == "aarch64" ]; then \
     else \
     ln -s /lib/ld-musl-x86_64.so.1 /bin/ldd; \
     fi
+# Set teh version here as otherwise its easy to invalidate the cache with a version change
```
jimmykarily:
```suggestion
# Set the version here as otherwise its easy to invalidate the cache with a version change
```

jimmykarily (approved):

mauromorales (approved):

Dockerfile
```diff
@@ -422,8 +422,8 @@ RUN for i in $(seq -w 1 ${PATCH_LEVEL}); do \
 WORKDIR /sources/downloads
 
 FROM sources-downloader-base AS libkcapi-download
-ARG LIBKKCAPI_VERSION=1.5.0
-RUN wget -q https://github.com/smuellerDD/libkcapi/archive/refs/tags/v${LIBKKCAPI_VERSION}.tar.gz -O libkcapi.tar.gz
+ARG LIBKCAPI_VERSION=1.5.0
```
Itxaka, line 425: wrong VAR was used for updates loool

PR #392, opened 2026-05-28
Ci improvements

jimmykarily:
Extracted from the work on AWS sev-snp (https://github.com/kairos-io/hadron/pull/393)

This makes sure that dependencies are resolved on the checked-in kernel config files. This is a step that happens when the kernel compilation happens anyway, but we may have options which we think are applied but are removed because their deps are missing.


files/kernel/cloud.config
```diff
@@ -1,18 +1,18 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86_64 7.0.6 Kernel Configuration
+# Linux/x86_64 7.0.10 Kernel Configuration
 #
-CONFIG_CC_VERSION_TEXT="gcc (GCC) 16.1.1 20260430"
+CONFIG_CC_VERSION_TEXT="gcc (Alpine 15.2.0) 15.2.0"
```
mudler, line 5: seem an oversight? I guess you wanted only to commit CI improvements here?

jimmykarily: The improvement in this case is that we run `olddefconfig` on PRs to ensure that the committed kernel configs have been "resolved". And since this PR introduces that, I had to run `olddefconfig` myself to make it pass. That's how this was introduced.

Regarding GCC vs Alpine in this line, see my comment here: https://github.com/kairos-io/hadron/pull/393#discussion_r3317964204

The new make target runs `olddefconfig` in a container using different tooling than that used in the final build so it changes some additional (informational) lines. Initially I added a regexp to exclude those changes but then I thought that even the `(GCC) 16.1.1 ...` line, was manually committed once and nothing guarantees that this is still the thing applied at build time (there is no automation to commit this back). So I thought, it doesn't matter if we keep the values from the helper make target instead.

jimmykarily: All the config changes you see in this PR were introduced by simply running `olddefconfig` on the existing configs. No other changes were introduced. This version of the configs is the one applied at build time, we just never committed it back.

jimmykarily:
Merged as part of: https://github.com/kairos-io/hadron/pull/393

closing.

PR #406, opened 2026-06-04
ci(riscv64): split PR pipeline and improve Quay cache reuse

mauromorales:
## Summary

- Re-enable `pull_request` on the RISC-V PR workflow and split the slow final image into `hadron-grub-warm` (cache-only) and `hadron-grub` (push to ttl.sh).
- Build **cloud** and **default** kernels in parallel with separate Quay cache keys (`kernel-misc-{cloud,default}-riscv64`).
- Write registry cache on every stage during PR runs (requires Quay secrets).
- Remove Kairos init and ISO jobs from PR CI so we can validate build timing before changing the main release workflow.

## Motivation

The v0.3.1 release `hadron-bios-riscv64` job hit the 6h GitHub Actions limit. Root causes included a single monolithic `target: default` build, kernel cache only warmed for `cloud`, and no `grub-*` cache populated before the final image step.

This PR is intentionally scoped to `.github/workflows/PR_riscv64.yml` only.

## Test plan

- [ ] Confirm **PR build and test images (riscv64)** runs on this PR
- [ ] Verify `kernel (cloud)` and `kernel (default)` complete in parallel
- [ ] Verify `hadron-grub-warm` completes under 6h and writes `grub-{cloud,default}-riscv64` cache
- [ ] Verify `hadron-grub` is mostly cache hits after warm jobs
- [ ] After green PR CI, follow up with the same structure in `build-multiarch-images.yml`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/PR_riscv64.yml
+++ b/.github/workflows/PR_riscv64.yml
@@ -2,30 +2,26 @@
 name: 'PR build and test images (riscv64)'
 
 on:
-  #pull_request:
+  pull_request:
   workflow_dispatch:
 
 concurrency:
   group: ci-image-riscv64-${{ github.head_ref || github.ref }}-${{ github.repository }}
   cancel-in-progress: true
 
-# Everything runs on oracle-vm-32cpu-128gb-x86-64 with QEMU RISC-V emulation.
-# The Dockerfile has no FROM --platform overrides, so every stage executes riscv64
-# binaries regardless of runner.  QEMU on 32 cores + 128 GB beats native RISC-V on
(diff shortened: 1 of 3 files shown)


jimmykarily (approved):

Itxaka:
I think cache save on pr is a but useless. Either you overwrite the main branch cache which is bad as the base cache should be the main branch one, or have a branch based cache, which adds over time unless you delete or auto delete. 

Originally our pr pipelines also has cache but it was useless and kept overwriting the good caches so I disabled them. Not sure how to.work around this. Maybe GitHub cache for PR only? But that exhausts really fast between prs somotmgets also messy

Itxaka (approved):

PR #423, opened 2026-06-12
Use openssl rehash instead of c_rehash in certhash hook

Itxaka:
c_rehash is a perl script, so update-ca-certificates pulled in perl at runtime. openssl rehash is a builtin in the modern OpenSSL shipped in every Hadron image and produces the same c_rehash-style hash symlinks, so swapping to it drops the perl runtime dependency.

Also run openssl rehash at build time as a fast-fail verification that the command works in the image. The bundle is the only file present so it is skipped (no symlinks created); real hash symlinks are created at runtime when individual CA files land in /etc/ssl/certs.


jimmykarily (approved):

tests/acceptance_test.go
jimmykarily: This should fail now?

Itxaka: Yes, im dumb and wondered why It was failing on a run parts test. Yikes


@copilot fix this

PR #441, opened 2026-06-18
feat: add conntrack-tools to the final image

Itxaka:
Adds conntrack-tools (conntrack + conntrackd) and its netfilter runtime
libraries (libnfnetlink, libnetfilter_conntrack/cttimeout/cthelper/queue)
to the full image, alongside iptables. Needed for k8s/kube-proxy.

Kept out of the base container target on purpose: that image targets
docker-style runtimes, not baremetal, so it does not get the full tools.
The conntrack-tools build stage is placed after libtirpc since conntrackd's
RPC sync support requires it at configure time. New source downloads are
appended at the end of the sources-downloader COPY block to avoid
invalidating the kernel build cache.


.github/scripts/image-size-report.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
+# Compare the full Hadron images built for this PR against the published
+# ":main" baseline and emit a Markdown size report to stdout.
+#
+# Used by the "image-size-report" job in PR_amd64.yml so reviewers can see how
+# much a change (e.g. adding a new dependency) grows the shipped images.
+#
+# Usage: image-size-report.sh <pr-sha> <repo-slug>
+#   <pr-sha>    git SHA the PR images were tagged with (github.sha)
```
mudler, line 9: nice!

mudler (approved):

PR #478, opened 2026-06-25
attach a report of package versions in the release notes

mudler:
This PR adds scaffolding to generate versions of packages included in the hadron image during build, and it also generates a page in gh-page that is front-facing showing the current versions in the image.

### DIFF
--- a/.github/workflows/build-multiarch-images.yml
+++ b/.github/workflows/build-multiarch-images.yml
@@ -1403,3 +1403,44 @@ jobs:
           done
           echo "Could not push history after retries" >&2
           exit 1
+
+  release-notes:
+    name: Update release notes with component manifest
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/v')
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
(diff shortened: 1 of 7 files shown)


Itxaka (approved):
I want this to be run inside the hadronnimage and shipped, sonwe can start adding support for trivy and whatnot to scan the images and report the packages. Would It be possible to add It as part of the container and final image under a stable path?


Otherwise what i thougth was every built software would add itself into a well know route as a $Name.version file which contains the versión number on It and then in the final step merge everything into a Ini file (easier to parse in bash) or json file(nicer for external compsumtion) so we always ship the versión as part of the image

mudler:
> I want this to be run inside the hadronnimage and shipped, sonwe can start adding support for trivy and whatnot to scan the images and report the packages. Would It be possible to add It as part of the container and final image under a stable path?

I guess so, yup, this one was a quick win by scanning the Dockerfile (given we are consistent in the arg naming PROGRAM_VERSION). It's dead simple to change the logic as this is a bunch of scripts.

> 
> Otherwise what i thougth was every built software would add itself into a well know route as a $Name.version file which contains the versión number on It and then in the final step merge everything into a Ini file (easier to parse in bash) or json file(nicer for external compsumtion) so we always ship the versión as part of the image

I'm fine either way, I went really KISS here and parsed rather than building file/layering. These sounds all good follow-ups

PR #480, opened 2026-06-26
feat(pam): STIG password complexity, history and last-login (#4058 finding 6)

Itxaka:
## Summary

Implements **PAM password-quality hardening** from the STIG spike (kairos-io/kairos#4058, finding 6) for the full Hadron image. Follow-up to #474 (which shipped findings 1/2/4/8).

### New build packages (full image only — the minimal `container` target has no PAM)
- **cracklib 2.10.3** — dictionary backend. `pw_dict` is generated at build time from the official wordlist (~1.9M words) with the freshly built `cracklib-packer` (runs natively or under binfmt on every arch; the 2.10 format is byte-order-aware → valid on all little-endian targets). We do **not** use `create-cracklib-dict`/`cracklib-format`: those assume GNU `gzip`/`tr` and silently produce a ~14k-word garbage dict under busybox (busybox `gzip -cdf` errors on non-gzip input; `tr -cd '[:graph:]'` treats the class as literal chars). Instead the list is formatted with busybox-safe operators and piped straight to `cracklib-packer`, with an `ls` guard so an empty dict fails the build.
- **libpwquality 1.4.5** — `pam_pwquality.so` + `pwscore`, built `--enable-pam` against cracklib, `securedir=/usr/lib/security`.
- **lastlog2 1.2.0** — `pam_lastlog2.so` + `lastlog2` CLI (`pam_lastlog` was removed upstream in linux-PAM 1.6). sqlite-backed db at `/var/lib/lastlog/lastlog2.db`, created by its shipped `tmpfiles.d` entry. `libsqlite3` was previously build-only, so the shared lib is now shipped into the full image so the module can load.

### PAM stack / config
- `files/pam/system-auth`: add `pam_pwquality` (`retry=3 enforce_for_root`) and `pam_pwhistory` (`remember=5 use_authtok`) before `pam_unix`, and **drop `nullok`** (no empty passwords).
- `files/pam/system-login`: add `pam_lastlog2.so showfailed` to the session stack.
- `files/security/pwquality.conf`: RHEL 9 STIG values (minlen 15, one of each class, minclass 4, maxrepeat 3, maxclassrepeat 4, difok 8, dictcheck, gecoscheck, `enforce_for_root`).

### Provisioning safety
This only gates **interactive** password changes. Kairos/yip provisions by writing the `/etc/shadow` hash directly (`mudler/yip` `pkg/plugins/user.go` — no `chpasswd`/PAM), and login uses the pam_unix **auth** stack (hash compare), so weak cloud-init passwords still provision and `enforce_for_root` is safe to ship. Existing test credentials are unchanged.

## Testing

- **Built the full amd64 image locally** (`make build-hadron ARCH=amd64`) and verified against it: `pwscore` rejects weak / accepts strong (dict loads, ~1.9M-word dict shipped), modules + libs resolve (`pam_pwquality→libpwquality→libcrack→libz`; `pam_lastlog2→liblastlog2→libsqlite3`).
- **`image-structure` suite: 26 passed / 0 failed** against the built image (new pwquality + lastlog2 specs run, gated by `skipUnlessFullImage`).
- New VM assertions `assertPamPasswordPolicy` + `assertLastlog2` wired into the `acceptance` and `uki` specs (modules present, hardened `pwquality.conf`, wired into `system-auth` without `nullok`, `pwscore` reject/accept; lastlog2 records the SSH login). These run on the QEMU acceptance jobs in CI (amd64/arm64/riscv64 + FIPS).

Part of kairos-io/kairos#4058.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -368,6 +368,19 @@ FROM sources-downloader-base AS pam-download
 ARG PAM_VERSION=1.7.2
 RUN wget -q https://github.com/linux-pam/linux-pam/releases/download/v${PAM_VERSION}/Linux-PAM-${PAM_VERSION}.tar.xz -O pam.tar.xz
 
+FROM sources-downloader-base AS cracklib-download
+ARG CRACKLIB_VERSION=2.10.3
+RUN wget -q https://github.com/cracklib/cracklib/releases/download/v${CRACKLIB_VERSION}/cracklib-${CRACKLIB_VERSION}.tar.xz -O cracklib.tar.xz
+RUN wget -q https://github.com/cracklib/cracklib/releases/download/v${CRACKLIB_VERSION}/cracklib-words-${CRACKLIB_VERSION}.bz2 -O cracklib-words.bz2
+
+FROM sources-downloader-base AS libpwquality-download
+ARG LIBPWQUALITY_VERSION=1.4.5
+RUN wget -q https://github.com/libpwquality/libpwquality/releases/download/libpwquality-${LIBPWQUALITY_VERSION}/libpwquality-${LIBPWQUALITY_VERSION}.tar.bz2 -O libpwquality.tar.bz2
+
+FROM sources-downloader-base AS lastlog2-download
+ARG LASTLOG2_VERSION=1.2.0
+RUN wget -q https://github.com/thkukuk/lastlog2/releases/download/v${LASTLOG2_VERSION}/lastlog2-${LASTLOG2_VERSION}.tar.xz -O lastlog2.tar.xz
+
 FROM sources-downloader-base AS shadow-download
 ARG SHADOW_VERSION=4.19.4
 RUN wget -q https://github.com/shadow-maint/shadow/releases/download/${SHADOW_VERSION}/shadow-${SHADOW_VERSION}.tar.xz -O shadow.tar.xz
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1807,6 +1824,83 @@ RUN DESTDIR=/pam ninja -j${JOBS} -C buildDir install
 COPY files/pam/* /pam/etc/pam.d/
 
 
+# cracklib: dictionary-based password checking, dependency of libpwquality (STIG #6)
+FROM rsync AS cracklib
```
mudler, line 1828: I'm not sure to be fair if this should be part of hadron core. I mean, I do get it that one can have interactive password changes, but hadron is not really made for interactive mode, and indeed logins are or either written in the cloud config (if existing), everything else won't persist, what are the benefits here? 

Users could add these additionally in their images, no?

Itxaka:
Agreed — Hadron is immutable and non-interactive: logins come from the cloud config and interactive password changes aren't really a use case here, so the PAM password-complexity controls (finding #6) make more sense applied downstream in consumer images than baked into core. Closing. Can revisit as a docs/example pattern under `examples/` if that's useful for STIG-minded consumers.

PR #503, opened 2026-07-07
feat: unify firmware and layers snapshots into /components/ page

Itxaka:
Add Firmware and Layers tabs to the existing /components/ dashboard so users get a single view of Hadron versions, firmware releases, and layer images.

- Extract the inline heredoc from `gen-snapshot.sh` into `hack/components-template.html`. The template gains hash routing (`#components`, `#firmware`, `#layers[/<name>[/<tag>]]`) while keeping back-compat for the legacy bare-ref hash
- Add `hack/gen-external.sh` to fetch `data.json` from hadron-firmware and `releases.json` from hadron-layers into `docs/static/components/`. Validates JSON before atomic rename, must run after `gen-snapshot.sh` since that wipes the dir
- Wire the fetch step into `.github/workflows/gh-pages.yml` right after the snapshot step and before the Docusaurus build. Refresh cadence rides on the existing triggers (main pushes, tags, autobumper PRs) so no cron needed
- Add Components/Firmware/Layers navbar entries in `docusaurus.config.js` using `pathname://` to bypass the broken-link checker, since `/components/` is a hand-authored static file

Firmware tab surfaces the release dropdown, per-component image ref with copy button, sysext download link, filter box and a stale banner when the snapshot is older than 7 days. Layers tab renders one card per layer with latest version, description, image ref and expandable tag history.


Itxaka:
improved a bit:

image

image


image


image

mudler (approved):
Nice!

PR #523, opened 2026-07-16
Add getent and nslookup

mudler:
getent is commonly used in scripts and would silently break, nslooup similarly is widely used and should be safe to add

```
  useradd[1468]: failed adding user 'agent', exit code: 4
  kairos-agent: sh: line 4: getent: command not found
  kairos-agent: useradd: UID 1000 is not unique
```

found it in https://github.com/kairos-community/hadron-desktop/pull/8


Itxaka (approved):

tests/image_structure_test.go
```diff
@@ -191,8 +191,20 @@ var _ = Describe("hadron container image structure", Label("image-structure"), f
 		Entry("rsync", "rsync", "--version"),
 		Entry("grep", "grep", "--version"),
 		Entry("find", "find", "--version"),
+		Entry("getent", "getent", "passwd", "root"),
```
Itxaka, line 194: can we add nslookup as well?

Itxaka: @copilot add nslookup test here

Itxaka:
This passed the tests and the build so Im going to merge this already no need to spend cpu cycles

PR #526, opened 2026-07-21
Automatic version bumps

ci-robbot:
No version metadata was collected for this update.


Itxaka:
I guess this is due to the default kernel values or something, we dont have them properly?

PR #545, opened 2026-07-31
fix(grub): add the sleep module to the EFI grub image

Itxaka:
AuroraBoot's livecd grub config uses `sleep --verbose --interruptible 30`
to give its boot-to-RAM entry a cancellable countdown. On BIOS that works
because AuroraBoot ships its own eltorito.img with the sleep module built
in, but our EFI images were built without it, so booting the same config
under EFI failed with "can't find command `sleep`" and dumped you back at
the grub prompt.

One-word fix: `sleep` added to the grub-mkimage module list in the
grub-efi Dockerfile stage, which covers grubx64.efi, grubaa64.efi and
grubriscv64.efi.


Itxaka:
Dropping this: the boot-to-RAM menu entry in AuroraBoot is being reworked to not depend on the sleep module at all, so no grub changes needed here.

PR #554, opened 2026-08-05
Automatic version bumps for kernel

ci-robbot:
## Summary

- kernel was updated 7.1.5 -> 7.1.8

### kernel
- Version: `7.1.5` -> `7.1.8`
- Changelog: [kernel](https://www.kernel.org/category/releases.html)

PR #563, opened 2026-08-13
Automatic bumps for kernel

mudler-agent:
Replaces #554 with the kernel bump rebased onto current `main`.

This updates the kernel from 7.1.5 to 7.1.8, records the new tarball checksum in `sources.yaml`, and carries forward the six regenerated kernel configs from #554.

Verification:

- kernel version checks for all six configs
- BTF checks for the four enabled architecture configs
- `make test-components`
- `git diff --check`
- config diff matches #554 byte-for-byte


mudler-agent:
The branch is rebased on current `main` (0 commits behind), mergeable, and DCO passes. Both fork-triggered build workflows ended as `action_required` without creating jobs, so a maintainer needs to approve the workflow runs before CI can execute: [riscv64](https://github.com/kairos-io/hadron/actions/runs/31687390615) and [multiarch](https://github.com/kairos-io/hadron/actions/runs/31687390630). No branch change is needed.

mudler (approved):

mudler:
@mudler-agent fix CI issues

mudler-agent:
CI root cause: both `toolchain-amd64` and `toolchain-arm64` fail before compilation because `ghcr.io/kairos-io/hadron-sources/linux:7.1.8` does not exist. The reusable `populate-sources` workflow detects the missing tag, but intentionally skips publishing for fork PRs. This PR is from `mudler-agent/hadron`, so the toolchain jobs cannot proceed. A maintainer-owned branch containing this `sources.yaml` bump must run `populate-sources.yml` once to publish the verified multi-arch source image, then the failed PR jobs can be rerun. No PR code change can safely fix this without bypassing the source-cache security model.

PR #569, opened 2026-08-19
ci: pin the kairos Dockerfile fetch to a commit, not master

mauro-agent:
doing this while the problem is on kairos master, we can remove later but I think it's better to have a pinned version than to be guessing what new thing broke


mauro-agent:
This PR's own \`acceptance tests\` job will keep failing on \`permitrootlogin prohibit-password\` until [#570](https://github.com/kairos-io/hadron/pull/570) merges — that's the STIG-conflict fix. Pinning to a reproducible commit here doesn't change it: \`v0.17.3\` is the referenced kairos-init release either way, so the conflict is present regardless of which kairos SHA this pins to.

Neither PR will show fully green in isolation right now — #570 fixes the failing test, this one fixes the reproducibility. Both need to merge.

mauro-agent:
Combined into #571 so CI can actually go green on one PR — this one alone couldn't (the STIG conflict is present regardless of the pin value). Closing in favor of it.

PR #570, opened 2026-08-20
fix(ssh): renumber Hadron's sshd drop-ins to sort before kairos-init's

mauro-agent:
## Why

sshd applies `sshd_config.d/*.conf` in lexical order and uses the **first** value it sees per directive. `kairos-init` v0.17.3 added its own hardening drop-in at `/etc/ssh/sshd_config.d/05-kairos-hardening.conf`. Hadron's three sshd drop-ins were numbered `99-hadron-stig.conf`, `100-hadron-crypto.conf` and `100-hadron-fips.conf` — all sorting *after* `05-`, so kairos-init's values silently win on every directive both sides set.

Two different severities:

- `PermitRootLogin` and the rest of `99-hadron-stig.conf`'s access/session/login directives losing to kairos-init's defaults broke hadron's own acceptance tests: `expected effective sshd config to contain "permitrootlogin prohibit-password"`, got `permitrootlogin no`.
- `Ciphers`/`KexAlgorithms`/`MACs`/`HostKeyAlgorithms` losing to kairos-init's DevSec-baseline list in `100-hadron-fips.conf` is a **real compliance break, not a cosmetic mismatch**: FIPS images were silently running a non-FIPS-validated crypto set instead of the pinned FIPS-approved one. Nothing in CI caught this — the FIPS-crypto-isolation test only checks that the STIG file carries no crypto keywords, not that the crypto file's own values are actually the ones in effect.

## What changed

Renamed:

| old | new |
|---|---|
| `99-hadron-stig.conf` | `01-hadron-stig.conf` |
| `100-hadron-crypto.conf` | `02-hadron-crypto.conf` |
| `100-hadron-fips.conf` | `02-hadron-fips.conf` |
| `99-hadron.conf` (generated inline in `Dockerfile.tmpl`, sets `UsePAM`/`PrintMotd`) | `03-hadron.conf` |

The last one doesn't currently collide on a value that matters (both sides set `UsePAM yes`), included for consistency so no Hadron sshd drop-in relies on winning by accident.

All three still sort relative to each other exactly as before (01 before 02 before 03; the two crypto files are still mutually exclusive per image), just now uniformly below kairos-init's `05-`. Updated the one test that reads a STIG file by path (`tests/image_structure_test.go`), and the file headers / `Dockerfile.tmpl` comments that explained the old `99-`/`100-` reasoning — the new comments explain the actual bug and cite the acceptance-test failure it caused, so a future reader doesn't have to rediscover this.

## Relationship to #569

**This is the fix; [#569](https://github.com/kairos-io/hadron/pull/569) is a separate, narrower change (pinning the unpinned kairos Dockerfile fetch) that surfaced this bug but does not fix it.** #569's own CI will keep failing the `permitrootlogin prohibit-password` acceptance check until this PR merges too — pinning to a reproducible commit doesn't change *which* kairos-init release that commit points at (`v0.17.3` either way), so the STIG conflict is present regardless of the pin value. Both need to merge for either to go fully green; noted the dependency on #569 as well.

## Testing

- `./hack/render.sh` confirms `Dockerfile` (gitignored, generated) picks up the renamed paths cleanly, no other tracked-file drift.
- `go vet ./...` on `tests/` is clean.
- Did not boot-test locally (needs the actual image build + VM boot cycle); relying on CI's `acceptance tests` job, which is exactly the one currently red for this reason.


mauro-agent:
Combined into #571 along with #569 so CI can actually go green on one PR — this fix needed the reproducible pin to verify against. Closing in favor of it.

PR #573, opened 2026-08-21
ci(riscv64): lift the populate-sources/stage0 gate for one timing run

mauro-agent:
Follow-up to kairos-io/hadron#572 (merged) and the un-split riscv64 pipeline. Lifts the `if: ${{ false }}` gate on `populate-sources` and `stage0` in `PR_riscv64.yml` only, unblocking the full `stage0 -> stage1 -> rsync -> ... -> hadron-bios` chain for this PR's own CI run.

**Purpose**: get one real CI run of the un-split, x86+QEMU-only riscv64 pipeline, and a warm-cache timing number — the evidence needed before proposing re-enabling riscv64 CI to kairos-io maintainers.

**Scope, deliberately narrow**:
- `check-kernel-configs` stays gated off; it's unrelated to the build chain.
- `build-multiarch-images.yml`'s riscv64 jobs (release path) are untouched.
- This is a one-shot measurement, not a decision to keep riscv64 CI on. Revert after reading the timing, or open a follow-up to remove the gates for good if the number looks good.

Draft, per repo convention; not proposing merge yet.

### DIFF
--- a/.github/workflows/PR_riscv64.yml
+++ b/.github/workflows/PR_riscv64.yml
@@ -25,8 +25,24 @@ concurrency:
 # comment). hadron-bios now builds the whole `default` target in one job, on
 # the same x86+QEMU runners as everything else in this pipeline — no OCI
 # hand-off, no native riscv64 runner.
+#
+# toolchain, container and hadron-bios build no runnable image artifact:
+# unlike PR_multiarch.yml's amd64/arm64 chains, nothing in this PR-only
+# pipeline consumes one downstream (no build-kairos-*, no ISO job), so there
+# is nothing to hand off as a tarball. Every stage here, from stage0 through
+# hadron-bios, is glued to the next purely by BuildKit's own registry-cache
+# reuse (cache-from/cache-to below) inside one shared Dockerfile with a
+# progressively deeper `target:` — matches PR_multiarch.yml's toolchain-amd64
+# job, which builds the same range in a single call with no CI-job boundary.
 
 env:
+  # Per-PR BuildKit cache. Each build job reads its own pr-<number>-* tag first
+  # (warm from a prior push or an earlier job in this same run) and falls back
+  # to the shared main cache tag. Only a same-repo PR can WRITE it — a fork PR
+  # has no secrets, so it skips the quay.io login and cache-to, and still
+  # reads the public cache-from. Matches PR_multiarch.yml's PR_CACHE_WRITE.
+  PR_CACHE_WRITE: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+  PR_NUM: ${{ github.event.pull_request.number }}
   CACHE_TO: mode=max,ignore-error=true,compression=zstd,compression-level=3,oci-mediatypes=true,image-manifest=true
   CACHE_TO_MIN: mode=min,ignore-error=true,compression=zstd,compression-level=3,oci-mediatypes=true,image-manifest=true
 
@@ -93,7 +109,8 @@ jobs:
           buildkitd-config-inline: |
             [worker.oci]
               max-parallelism = 4
-      - name: Log in to quay.io for cache
+      - name: Log in to quay.io (push per-PR build cache)
+        if: ${{ env.PR_CACHE_WRITE == 'true' }}
         uses: docker/login-action@v4
         with:
           registry: quay.io
@@ -114,8 +131,9 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           target: stage0
           cache-from: |
+            type=registry,ref=quay.io/kairos/hadron-build-cache:pr-${{ env.PR_NUM }}-stage0-riscv64,mode=max
             type=registry,ref=quay.io/kairos/hadron-build-cache:stage0-riscv64,mode=max
-          cache-to: type=registry,ref=quay.io/kairos/hadron-build-cache:stage0-riscv64,${{ env.CACHE_TO }}
+          cache-to: ${{ env.PR_CACHE_WRITE == 'true' && format('type=registry,ref=quay.io/kairos/hadron-build-cache:pr-{0}-stage0-riscv64,{1}', env.PR_NUM, env.CACHE_TO) || '' }}
           build-args: |
             ARCH=riscv64
             BUILD_ARCH=riscv64
@@ -137,7 +155,8 @@ jobs:
           buildkitd-config-inline: |
             [worker.oci]
               max-parallelism = 4
-      - name: Log in to quay.io for cache
+      - name: Log in to quay.io (push per-PR build cache)
+        if: ${{ env.PR_CACHE_WRITE == 'true' }}
         uses: docker/login-action@v4
         with:
           registry: quay.io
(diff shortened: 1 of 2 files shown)


mauro-agent:
Pushed `ab45986`: mirrors `musl.libc.org` for mussel's musl fetch, same pattern as the existing GNU mirror redirect two lines above. This has now failed the exact same way on three separate runs since 21 August (this branch twice, and #574), consistent with the DDoS/rate-limiting incident noted in e1d415d rather than one-off bad luck.

Redirects to Alpine's own distfiles cache (`distfiles.alpinelinux.org/distfiles/v3.20`), which mirrors the identical upstream tarball. Verified byte-for-byte against mussel's pinned sha256 for musl-1.2.5 before pushing.

mauro-agent:
Follow-up on `ab45986`: the musl.libc.org mirror fix worked — this run got past `stage0` cleanly for the first time. New problem surfaced downstream in `stage1`: it stalled for ~38 minutes after the musl-patch step (last log line at 16:56:48, cancelled at 17:34:58 with `The operation was canceled`), no error before the cancellation, no obvious cause in the log. [Run 32646114617](https://github.com/kairos-io/hadron/actions/runs/32646114617). Separate issue from the musl fetch, needs its own look — didn't dig further, flagging for whoever picks this back up.

mauro-agent:
Real full-chain timing data from this branch, run [32739261752](https://github.com/kairos-io/hadron/actions/runs/32739261752), 2026-08-24 14:31 UTC to 2026-08-25 07:41 UTC:

- `stage0`: 2h10m, success
- `stage1`: 4h9m, success
- `rsync`: 4h43m, success
- `kernel (cloud)`, `kernel (default)`, `systemd`, `python`, `cmake`: all started around 01:40 UTC. All were cancelled at 6h00m22s, exactly at GitHub's hard per-job execution limit for GitHub-hosted runners.
- `toolchain`, `container`, `hadron-bios`: never ran. Their inputs did not finish.

Total wall time to this point: about 17 hours. Four build stages did not finish, and the pipeline never reached its final stages.

This is not a workflow bug and not a flake. GitHub caps every job on a GitHub-hosted runner at 6 hours of execution time. That cap is separate from `timeout-minutes` and cannot be raised on hosted runners (self-hosted runners allow up to 5 days per job).

Full analysis and the recommendation this leads to are on the tracking ticket, [mauromorales/mission-control#113](https://github.com/mauromorales/mission-control/issues/113) (private repo).

mauromorales:
@mauro-agent re-ran the failed stages, keep an eye on this and ping me when it's done

mauro-agent:
⏳ **Work claimed by `mauro-agent`**, an AI agent operated by `@mauromorales`, at 2026-09-07T15:31:31Z. If you are another lane about to start substantial work here, please check for a follow-up comment before duplicating it -- this marker exists specifically to prevent that (see mauromorales/mission-control#589).

mauro-agent:
**Unattended agent review/monitoring notice** (mauro-agent, ADR-009 decision 2): this comment is from an unattended agent. Mauro has not read this pull request.

Status check for `@mauromorales`: 2 checks still running (`toolchain`, `toolchain-arm64`), all others pass. Will follow up here when all checks finish.

mauro-agent:
🤖 **Automated review by `mauro-agent`** — an AI agent operated by `@mauromorales`. This is not Mauro's personal review; he has not read this PR unless he says so himself. If you want a human, feel free to ping him directly.

CI status update. 41 of 42 checks pass, `toolchain` and `toolchain-arm64` are now green.

One check remains: `container`. Its "Build container" step has been running since 2026-09-07T23:58:23Z, over 6 hours. Other container jobs on this run (`container-amd64`, `container-arm64`) finished in 36-51 minutes. This looks stuck, not just slow.

Job: https://github.com/kairos-io/hadron/actions/runs/33027575710/job/101892810405

mauro-agent:
## Why this run is so slow: the registry cache is ~3 months stale, and every job rebuilds the whole chain from scratch

Checked the riscv64 cache tags on quay.io/kairos/hadron-build-cache. All last written between 2026-06-04 and 2026-06-12 — right when #419 disabled this pipeline. Nothing has refreshed them since: this PR runs from a fork, and `PR_CACHE_WRITE` only writes cache for same-repo PRs, so every run here has read-only access to a cache that hasn't moved in three months.

The `toolchain` job (rerun, this run) is the clearest evidence. Its log shows 1228 `DONE` build steps against only 53 `CACHED` — a ~4% hit rate. The two longest single steps inside that one job:

| Step | Duration | What it is |
|---|---|---|
| `#339 [stage0 9/9] RUN ./mussel riscv64 ...` | 2h9m | The mussel cross-compiler build -- this is `stage0`, re-run from scratch |
| `#407 [gcc-stage0 14/15] RUN make ... -j16` | 1h50m | GCC build inside stage0 |

Those two steps alone are 4h -- roughly half the job's 8h37m -- and they're `stage0` work, not `toolchain` work. Because `--target toolchain` needs the whole dependency chain and none of it is coming from cache, this one job is quietly rebuilding `stage0` through `toolchain` end to end, on its own, independent of the `stage0`/`stage1`/`rsync`/etc. jobs that already ran earlier in this same workflow run. `container` (still running as I write this, 6h48m+) is almost certainly doing the same thing again.

So the slowness isn't the self-hosted runner or the pipeline design -- it's that a stale cache turns every downstream job into a full rebuild of everything before it, and with N jobs each needing the chain, that's N redundant full rebuilds instead of one.

**The catch:** this can't self-heal. A fork PR will never write to this cache, so it stays stale until a same-repo run builds and pushes to it. Options once riscv64 CI comes back for real: a same-repo branch/push to warm the cache once, or accept the first same-repo run pays the cold-cache cost and everything after benefits.

mauromorales:
@mauro-agent this will be a problem if every external contribution starts cold, IMO we need a solution where external contribution can also benefit from the cache

mauro-agent:
⏳ **Work claimed by `mauro-agent`**, an AI agent operated by `@mauromorales`, at 2026-09-08T07:54:12Z. If you are another lane about to start substantial work here, please check for a follow-up comment before duplicating it -- this marker exists specifically to prevent that (see mauromorales/mission-control#589).

mauro-agent:
🤖 **Automated review by `mauro-agent`** — an AI agent operated by `@mauromorales`. This is not Mauro's personal review; he has not read this PR unless he says so himself. If you want a human, feel free to ping him directly.

On the cache question: this is not actually fork-specific. `PR_multiarch.yml` already has the same restriction for amd64/arm64 (`PR_CACHE_WRITE` gates cache *writes* on same-repo PRs only, because fork PRs never get repo secrets, so they cannot authenticate to quay.io to push). Every PR, fork or not, still *reads* the shared cache via `cache-from`. That has never been a problem for amd64/arm64 in practice, because `build-multiarch-images.yml` rebuilds and re-pushes those cache layers on every push to `main`, so the shared cache stays warm on its own.

riscv64 uses the exact same pattern in `build-multiarch-images.yml` (`stage0-riscv64`, `stage1-riscv64`, etc. all write to `quay.io/kairos/hadron-build-cache` when `github.ref_name == 'main'`). The reason it is stale is not the fork boundary. It is that the riscv64 PR pipeline (`PR_riscv64.yml`) has been gated off since #419, so no riscv64 work has merged to `main` in about three months, so nothing has triggered a refresh. Once riscv64 CI is back and riscv64 changes merge to `main` again, the existing mechanism re-warms the cache automatically, same as it already does for amd64/arm64. No design change needed for that part.

One real gap: unlike amd64/arm64, there is no scheduled job that refreshes the cache independent of `main` activity, so a long pause in riscv64 merges (like this one) goes stale with nothing to catch it. A periodic same-repo workflow (nightly or weekly, `workflow_dispatch` plus a cron trigger) that rebuilds and re-pushes the riscv64 cache layers would prevent this from recurring, without touching the fork write restriction at all.

For this PR specifically: since the shared riscv64 cache is stale right now, this run is paying a cold-start cost, not producing the warm-cache timing number the PR description was after. Re-running after a same-repo `main` build refreshes the riscv64 cache would give a more representative number.

Status: `hadron-bios (cloud)` and `hadron-bios (default)` are still in progress. Everything else passes.

mauro-agent:
hadron-bios (default) and hadron-bios (cloud) have been running 7h+ (started 06:53Z), same pattern as the container job earlier today that turned out stuck. Might be worth a re-run if these don't finish soon.

mauro-agent:
All checks are green now (44 pass, 3 skipped, 0 pending) -- the last stuck job, hadron-bios (default), finished at 8h51m.

PR #577, opened 2026-08-24
fix(stage0): pre-stage musl for mussel, stop hitting musl.libc.org

mauro-agent:
## Summary

`stage0` has been failing — confirmed from the actual failing log ([run 32481600204](https://github.com/kairos-io/hadron/actions/runs/32481600204/job/96769005748)), this isn't riscv64-specific at all:

```
curl: (28) Failed to connect to musl.libc.org:443 after 133527 ms: Could not connect to server
```

mussel (the vendored cross-toolchain builder) fetches musl straight from `musl.libc.org` with no fallback. `binutils`/`gcc`/`gmp`/`mpc`/`mpfr` already got a mirror-redirect fix for the same class of problem (`ftpmirror.gnu.org`'s rotator serving broken mirrors, see the existing comment above the `sed` in `Dockerfile.tmpl`) — musl never did.

## What this does

Confirmed from mussel's own source (`mpackage()` in [firasuke/mussel](https://github.com/firasuke/mussel)) that it skips its network fetch entirely when the tarball is already present at the exact local path it expects.

- Adds a `musl-toolchain` entry to `sources.yaml` — a **second, separately-versioned entry**, not a duplicate of the existing `musl` one (which is pinned to `1.2.6` for a different consumer). mussel's own vendored pin is `1.2.5`, confirmed by reading the pinned `MUSSEL_VERSION` commit directly, and the checksum used is mussel's own already-verified `musl_sum` value — not a new one I'm introducing.
- Cached via the existing per-package source cache (`populate-sources` / `ghcr.io/kairos-io/hadron-sources`, from #553) the same way every other source in this Dockerfile already is.
- Pre-staged into mussel's own source directory (`mussel/sources/musl/musl-1.2.5.tar.gz`) before `./mussel` runs, so it never touches `musl.libc.org` directly at all.

Rendered clean via `hack/render.sh` — no unresolved placeholders.

## Not done here

`populate-sources` needs a real run to actually publish `ghcr.io/kairos-io/hadron-sources/musl-toolchain:1.2.5` before this is exercised for real — the `musl-toolchain` entry in `sources.yaml` is picked up automatically by that workflow's own discovery step, no separate registration needed, but the image doesn't exist yet.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Dockerfile.tmpl
+++ b/Dockerfile.tmpl
@@ -32,6 +32,11 @@ RUN apk update && \
     curl m4 xz texinfo bison gawk gzip zstd-dev coreutils bzip2 tar rsync \
     git coreutils findutils pax-utils binutils
 
+# The tarball mussel itself needs for its cross-toolchain build (see
+# stage0 below), cached the same way every other source in this
+# Dockerfile is -- so mussel never has to fetch musl.libc.org directly.
+FROM ghcr.io/kairos-io/hadron-sources/musl-toolchain:${MUSL_TOOLCHAIN_VERSION} AS musl-toolchain-download
+
 FROM alpine-base AS stage0
 
 ########################################################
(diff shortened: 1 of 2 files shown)


mauro-agent:
Duplicate of #578, which is already approved.

PR #578, opened 2026-08-24
fix(stage0): pre-stage musl for mussel, stop hitting musl.libc.org

mauromorales:
stage0 has been failing (not riscv64-specific -- confirmed from the actual log) because mussel, the vendored cross-toolchain builder, fetches musl straight from musl.libc.org with no fallback, and that host has been intermittently unreachable. binutils/gcc/gmp/mpc/mpfr already got a mirror-redirect fix for the same class of problem (ftpmirror.gnu.org's rotator serving broken mirrors); musl never did.

Confirmed from mussel's own source (mpackage()) that it skips its network fetch entirely when the tarball is already present at the exact local path it expects. Adds a musl-toolchain entry to sources.yaml -- a second, separately-versioned entry, not a duplicate of the existing `musl` one, which is pinned to 1.2.6 for a different consumer. mussel's own vendored pin is 1.2.5, and mussel's own already-verified musl_sum checksum is what's used here, not a new value. Cached via the existing per-package source cache (populate- sources / ghcr.io/kairos-io/hadron-sources, from #553) the same way every other source in this Dockerfile already is, then copied into mussel's own source directory before it runs, so it never touches musl.libc.org directly.

Rendered clean via hack/render.sh, no unresolved placeholders.

Co-developed-by: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/sources.yaml
+++ b/sources.yaml
@@ -645,6 +645,24 @@ packages:
       - https://musl.libc.org/releases/musl-${version}.tar.gz
     filename: musl.tar.gz
 
+  # A second, separately-versioned entry, not a duplicate of `musl` above.
+  # This is what mussel (the vendored cross-toolchain builder in stage0,
+  # not the final rootfs) pins internally -- 1.2.5, not 1.2.6 -- and
+  # mussel's own musl_sum checksum, already the value every prior
+  # successful build has verified against. mussel fetches this straight
+  # from musl.libc.org itself with no fallback, which is what's actually
+  # failing (stage0 timing out on musl.libc.org, not on riscv64 at all).
+  # Cached here the same way as every other source in this file, then
+  # pre-staged into mussel's own source directory before it runs, so it
+  # never touches musl.libc.org directly. See stage0 in Dockerfile.tmpl.
+  musl-toolchain:
+    version: "1.2.5"
+    version_arg: MUSL_TOOLCHAIN_VERSION
+    sha256: a9a118bbe84d8764da0ea0d28b3ab3fae8477fc7e4085d90102b8596fc7c75e4
+    urls:
+      - https://musl.libc.org/releases/musl-${version}.tar.gz
+    filename: musl-toolchain.tar.gz
+
   musl-fts:
     version: "1.2.7"
     version_arg: FTS_VERSION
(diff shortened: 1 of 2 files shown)


sources.yaml
```diff
@@ -645,6 +645,24 @@ packages:
       - https://musl.libc.org/releases/musl-${version}.tar.gz
     filename: musl.tar.gz
 
+  # A second, separately-versioned entry, not a duplicate of `musl` above.
+  # This is what mussel (the vendored cross-toolchain builder in stage0,
+  # not the final rootfs) pins internally -- 1.2.5, not 1.2.6 -- and
```
jimmykarily, line 650: Was the use of a different version intentional? Is this the same sources but from a different mirror? If yes, then maybe we only need to add one more url above and use the same source everywhere.

mauro-agent: Intentional, and not a mirror -- it's a different version for a different consumer, not the same source served two ways.

- `musl-toolchain` (1.2.5) is what mussel needs to bootstrap its own cross-compiler in stage0. mussel hard-pins 1.2.5 internally and checks the tarball against its own `musl_sum` checksum. The file has to land at the exact path mussel expects (`mussel/sources/musl/musl-1.2.5.tar.gz`) for its `mpackage()` fetch-skip logic to kick in.
- `musl` (1.2.6) is the actual target libc that ships in the final rootfs, built later (`musl-stage0` / `MUSL` stages) using the toolchain mussel produces.

So these are two genuinely different musl releases at two different points in the build (toolchain bootstrap vs. target libc), not one source with two paths to it. Pointing mussel at 1.2.6 would fail its own checksum check; making that work would mean patching mussel's pinned version and checksum too, which is a toolchain upgrade, out of scope for a network-reliability fix.

One more thing worth flagging: this isn't actually a mirror fix like binutils/gcc/gmp/mpc/mpfr got. Both `musl` and `musl-toolchain` still list only `musl.libc.org` as their URL, no fallback host added. The fix is that `musl-toolchain` goes through the existing sources-cache-and-copy mechanism (`populate-sources` -> `ghcr.io/kairos-io/hadron-sources`) and gets pre-staged into mussel's own source directory before mussel runs, so mussel's direct fetch never happens. A second URL on the existing `musl:` entry wouldn't help mussel either way, since mussel doesn't consult `sources.yaml`'s URL list at all, it only checks whether the tarball is already sitting at its expected local path.

Not an agent, Mauro reviewed and okayed this reply.

jimmykarily (approved):

PR #558, opened 2026-08-09
Automatic version bumps for shadow

ci-robbot:
## Summary

- shadow was updated 4.20.0 -> 4.20.2

### shadow
- Version: `4.20.0` -> `4.20.2`
- Release: [shadow-maint/shadow 4.20.2](https://github.com/shadow-maint/shadow/releases/tag/4.20.2)
- Notes:

<details>
<summary>Release notes excerpt (truncated)</summary>

shadow-4.20.2

Regression fixes:

-  lib/: Add missing include.

Other changes:

-  Remove unused build flag

---

Binary tarballs produced and signed with:

```
$ ./autogen.sh 
$ make -j4
$ make dist
$ gpg --detach-sign --armor shadow-4.20.2.tar.gz 
$ gpg --detach-sign --armor shadow-4.20.2.tar.xz
```

Source tarballs produced and signed with:

```

</details>

- Compare: [4.20.0...4.20.2](https://github.com/shadow-maint/shadow/compare/4.20.0...4.20.2)

PR #559, opened 2026-08-09
Automatic version bumps for libffi, libkcapi

ci-robbot:
## Summary

- libffi was updated 3.7.1 -> 3.8.0
- libkcapi was updated 1.5.0 -> 1.5.1

### libffi
- Version: `3.7.1` -> `3.8.0`
- Release: [libffi/libffi 3.8.0](https://github.com/libffi/libffi/releases/tag/v3.8.0)
- Notes:

<details>
<summary>Release notes excerpt</summary>

## New features

- **`FFI_TYPE_VECTOR` (SIMD) type support** with libffi-computed layout, for aarch64 and x86-64 (#1000, closes #773).
- **`ffi_call_plan_size`** — report the total memory a reusable call plan owns, for embedders that account for the memory held by long-lived plans.
- **powerpc64 ELFv2 `_Complex long double`** support for both IBM-128 (double-double) and IEEE-128 formats (#1003, closes #1001).

## Fixes

- **powerpc64 big-endian ELFv2 closures** returning 5-, 6-, or 7-byte structs: missing return jump-table entries produced a wrong result and leaked a libffi code pointer.
- **ia64 return-value jump-table desync** after the `FFI_TYPE_LAST` bump, which corrupted small-struct and HFA returns.
- **powerpc Darwin closure returns** broken by #951 (#1002).
- **i386 FreeBSD/OpenBSD small structs** now returned in registers (1, 2, 4, or 8 bytes), matching the platform ABI and fixing a segfault on struct returns through `ffi_call` and closures.

## Performance

- **Cache the static trampoline "unsupported" verdict** on hosts whose page size exceeds the trampoline table mapping, avoiding redundant re-initialization on every closure allocation (e.g. 64K-page aarch64).

</details>

- Compare: [v3.7.1...v3.8.0](https://github.com/libffi/libffi/compare/v3.7.1...v3.8.0)

### libkcapi
- Version: `1.5.0` -> `1.5.1`
- Release: [smuellerDD/libkcapi 1.5.1](https://github.com/smuellerDD/libkcapi/releases/tag/v1.5.1)
- Notes:

<details>
<summary>Release notes excerpt</summary>

* remove tests of ansi_cprng
* apply change suggestions derived from Claude code
* kcapi-enc: add tag processing in decryption without ciphertext for completeness reason
* fix memory corruption (CVE-2026-71226)
* fix potential for an infinite loop (CVE-2026-71227)
* add safety measure to prevent IV reuse (CVE-2026-71225)

</details>

[...]

PR #584, opened 2026-09-04
deps: bump shadow to 4.20.2 and drop upstreamed musl patch

itxaka-agent:
Bumps shadow from 4.20.0 to 4.20.2 in `sources.yaml` and drops the local musl `<stdint.h>` backport, which is now upstream.

Shadow 4.20.2 already contains 0182cfc9 ("subid: Add <stdint.h> for uintmax_t") — the same fix we cherry-picked as `patches/0001-shadow-subid-stdint-musl.patch` from upstream ba4419c5. I confirmed this by extracting `lib/find_new_sub_uids.c` from the 4.20.2 tarball and checking that `#include <stdint.h>` is present. Left as-is, the patch would fail to apply on rebuild.

So this PR bundles the bump with the cleanup:

- `sources.yaml`: shadow 4.20.0 → 4.20.2, sha256 updated for the new tarball
- `patches/0001-shadow-subid-stdint-musl.patch`: deleted
- `Dockerfile.tmpl`: drop the `COPY` + `RUN patch -p1` lines and the "drop this patch when bumping past 4.20.0" comment blocks in both the base shadow stage and the systemd shadow stage

Supersedes #558, which only touched `sources.yaml` and would have broken both shadow stages on rebuild. Closes #558.


mauromorales (approved):

itxaka-agent:
> action taken here.

`check-kernel-configs` was failing on `make test-render`: "fork render did not generate the verified libkcapi download stage". I reproduced the same failure on a clean `upstream/main` (`ef7838f`) with none of this PR's changes applied, so it isn't from the shadow bump in this PR. It comes from `#559` (libkcapi `1.5.0` → `1.5.1` in `sources.yaml`) landing on `main` while `tests/render-fork-sources.sh` still asserted the old `1.5.0` URL and sha256 in its expected output.

There's already an open fix for this upstream, `#585`, which does the same thing. Since merge order between that PR and this one isn't guaranteed, I fixed it forward on this branch too: `a0a8725` updates the three hardcoded `1.5.0` literals in `tests/render-fork-sources.sh` to `1.5.1` (matching the pin already on this branch's `sources.yaml`). `make test-render` passes locally now. If `#585` merges first, this commit becomes a no-op on rebase rather than a conflict — same resulting text.

itxaka-agent:
> action taken here.

CI on this PR went red on `populate-sources / discover`
(`http.client.RemoteDisconnected: Remote end closed connection without response`
talking to ghcr.io's token endpoint while checking which `sources.yaml`
packages already have a published cache tag). That is a transient failure on
the registry side, not something in this diff — the shadow bump and the
libkcapi test-literal fix don't touch that code path at all.

I don't have admin rights on this repo, so `gh run rerun --failed` and a
manual `workflow_dispatch` of the same workflow both came back
`403 Must have admin rights to Repository`. Rebased the branch onto
`upstream/main` instead (picked up `ef7838f`, an unrelated `lvm2` bump) and
force-pushed to my fork branch. New head is `173cf348`. That gives a clean
`pull_request: synchronize` event so CI reruns from scratch. DCO is already
green; the rest of the checks are running.

PR #560, opened 2026-08-10
Automatic version bumps for busybox

ci-robbot:
## Summary

- busybox was updated 1.37.0 -> 1.38.0

### busybox
- Version: `1.37.0` -> `1.38.0`
- Changelog: [busybox](https://git.busybox.net/busybox/log/)
- Compare: [1_37_0...1_38_0](https://git.busybox.net/busybox/diff/?id=1_38_0&id2=1_37_0)

PR #586, opened 2026-09-04
deps: bump busybox to 1.38.0 and regenerate minimal.config

itxaka-agent:
Bumps busybox 1.37.0 → 1.38.0 in `sources.yaml` and regenerates `files/busybox/minimal.config` so the build stays non-interactive.

## Why

1.38.0 adds new Kconfig symbols the checked-in `minimal.config` did not answer:

- `FEATURE_VERSION`, `FEATURE_PATH_TRAVERSAL_PROTECTION`, `USE_BB_CRYPT_YES`, `FEATURE_FDISK_BLKSIZE`, `FEATURE_TELNETD_SELFTEST_DEBUG`
- New applets: `SHA384SUM`, `LSBLK`, `UUIDGEN`, `SSL_SERVER`, `VMSTAT`
- `TLS` (was `n`, upstream default flipped to `y`)
- HUSH extras: `HUSH_NEED_FOR_SPEED`, `HUSH_ALIAS`, `HUSH_FUNCTION_KEYWORD`

The Dockerfile runs `make silentoldconfig`, which does not consume stdin — as soon as one unanswered symbol appears it aborts with `Console input/output is redirected. Run 'make oldconfig' to update configuration.` (exit 2). Reproduced locally with the 1.38.0 tarball and the current `minimal.config`.

## Fix

Regenerate `files/busybox/minimal.config` by running `make oldconfig` against 1.38.0 with upstream defaults for every new symbol, and commit the result as the new source of truth. `silentoldconfig` in the Dockerfile then finds no unanswered symbols and stays silent — no `yes` pipe, no runtime interactivity dance.

The applied upstream defaults are visible in the config diff (LSBLK, UUIDGEN, SSL_SERVER, TLS, VMSTAT, SHA384SUM, FEATURE_VERSION flipped on). Reviewers who want a stricter minimal footprint can flip specific symbols off in a follow-up.

Supersedes #560 (which only bumped `sources.yaml` and would have broken the build on rebuild). Closes #560.


mauromorales (approved):

itxaka-agent:
> action taken here.

`check-kernel-configs` was failing on `make test-render`: "fork render did not generate the verified libkcapi download stage". This branch was cut from `ef7838f`, one commit behind `upstream/main` at the time -- `#585` (merged after this PR was opened) updated `tests/render-fork-sources.sh` to match the libkcapi pin that had already landed, and this branch never picked that fix up.

Rebased onto `upstream/main` (now on top of `cabb59c`) and force-pushed to my fork branch; no conflicts, the busybox files are untouched by `#585`. New head is `2e10055`. `make test-render` passes locally now. DCO is already green; the rest of the checks are running from the fresh `pull_request: synchronize` event.

itxaka-agent:
> action taken here.

CI was red on `container-amd64` and `container-arm64` after the last push (`2e10055`). Both failed identically: `No URL served the expected bytes for busybox-1.38.0`.

Root cause: fork PRs render the Dockerfile in upstream mode, which turns the `busybox-download` stage into a direct `wget` against the single URL in `sources.yaml` (`https://busybox.net/downloads/busybox-1.38.0.tar.bz2`). That host is not answering TLS handshakes right now — a live upstream outage, not a bad checksum. I confirmed this two ways: `curl -I` to busybox.net times out repeatedly while unrelated hosts (github.com, pypi.org, ftp.gnu.org) respond normally, and I fetched `https://sources.openwrt.org/busybox-1.38.0.tar.bz2` directly — its sha256 is an exact match for the pin already in `sources.yaml`.

Fix in `a60268f`: added `https://sources.openwrt.org/busybox-${version}.tar.bz2` as a second URL for busybox, same pattern already used elsewhere in this file for GNU/Savannah/Alpine mirrors. The pinned sha256 is unchanged. Also extended `tests/render-fork-sources.sh`'s `unreliable`-host list to include `busybox.net`, so a future version bump can't silently drop the fallback and land back here — verified locally that `make test-render` fails without the `sources.yaml` change and passes with it.

CI should re-run automatically on the new push.

PR #597, opened 2026-09-14
feature: add audit userspace tools

### LINKED ISSUE #4637: hadron: add auditd + audit userspace to the package set
Split from kairos-io/kairos#112 (see kairos-io/kairos#4626, kairos-io/kairos#4627, kairos-io/kairos#4628, kairos-io/kairos#4629 for the rest of the CIS work).

Add `auditd` and its userspace tools (`auditctl`, `augenrules`, `ausearch`, `aureport`) to the Hadron package set so kairos-init images built on top of Hadron can ship a working CIS **Section 4 — Logging and Auditing** baseline.

## Why Hadron specifically

Distro-based Kairos images (Ubuntu, openSUSE, RHEL, Rocky, Debian, Alpine) can pull auditd from the upstream package repositories. Hadron builds its own userland from source through the mussel toolchain, so nothing lands unless it is explicitly added to the build. Without this issue closed, the kairos-init CIS L1 issue (kairos-io/kairos#4626) can never satisfy Section 4 on Hadron-based images.

## Scope

- Add the `audit` userspace (upstream: https://github.com/linux-audit/audit-userspace) as a new build stage in the Hadron Dockerfile, cross-compiled through mussel for `amd64`, `arm64`, and `riscv64`.
- Ship the `auditd` binary, `auditctl`, `augenrules`, `ausearch`, `aureport`, and the `libaudit`/`libauparse` shared libraries in the final rootfs.
- Add an init unit (systemd or OpenRC depending on Hadron's current init) that starts `auditd` early enough to catch boot-time events.
- Ship a minimal `/etc/audit/auditd.conf` and an empty `/etc/audit/rules.d/` directory. The CIS baseline rules themselves land in kairos-io/kairos#4626 on the kairos-init side.
- Verify kernel audit support (`CONFIG_AUDIT=y`, `CONFIG_AUDITSYSCALL=y`) is enabled in the Hadron kernel config; enable it if missing.

## Prerequisites

- kairos-io/kairos#4629 — `/var/log/audit` must persist across A/B upgrades before auditd is actually useful. Ship the package first, wire persistence second, land the rules third.

## Non-goals

- The CIS baseline audit rules — those belong in kairos-io/kairos#4626, not here.
- Remote log forwarding (audispd plugins for `syslog`/`remote`) — separate feature request.
- Hardening `auditd.conf` beyond the upstream defaults — belongs in kairos-io/kairos#4626 alongside the rules.

## Definition of done

- `auditd --version` works on a booted Hadron-based Kairos image on all three architectures.
- The daemon starts at boot and writes to `/var/log/audit/audit.log`.
- Binary size impact is documented on the PR (mussel-compiled audit userspace is small, but should be measured).
- kairos-io/kairos#4626 can install its baseline rules on top without pulling in a second copy of the userland.

---

*Filed by [`itxaka-agent`](https://github.com/itxaka-agent) after @itxaka's follow-up on kairos-io/kairos#112. Approved by @itxaka before posting.*

Itxaka:
As part of the CIS hardenings, audit is a important part of the system.

Build and provide the proper userspace tools. Final size should not be grater than 1Mb.

Unfortunately a patch is needed for a clean compile. sent the patch upstream as well but for now lets carry it by ourselves.

Only tests were done by IA in this PR.

Fixes https://github.com/kairos-io/kairos/issues/4637


itxaka-agent:
> action taken here.

Taking this over for review. It touches the mussel cross-compile stage set (adds a new `audit-userspace` build stage across amd64/arm64/riscv64), carries an out-of-tree patch, and ships a new binary set into the image, so I'll pull the branch, read the Dockerfile stage against the upstream `audit-userspace` source, check the patch against what it's fixing, and verify the image-structure test actually asserts what the PR claims (binary presence, size budget). Will report back with a verdict.

itxaka-agent:
## End-to-end verification: PASS

Built `hadron:local` locally, layered Kairos, cut ISO with AuroraBoot, booted in QEMU/KVM headless, and verified audit userspace runs and writes to disk.

### Pipeline

1. `make build-hadron` → `hadron:local` (405 MB)
2. `make build-kairos` → `hadron-init:latest` (573 MB)
3. AuroraBoot `v0.21.0-alpha.4` → `kairos-hadron-v0.5.2-dirty-core-amd64-generic-v0.0.0.iso` (304 MB)
4. QEMU boot: `-accel kvm -cpu host -m 3000 -smp 2 -vga std -display none -qmp unix:...`

### `image-structure` audit specs

12 / 12 pass against `hadron:local`:

```
Ran 12 of 49 Specs in 16.292 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 37 Skipped
```

Covers: `auditctl`, `ausearch`, `aureport`, `augenrules`, `autrace`, `audisp-syslog` crash-signal check; `/sbin/auditd` present; `/etc/audit/auditd.conf` contains `log_file`; `/etc/audit/rules.d` real directory; `/usr/lib/systemd/system/auditd.service` has `ExecStart=`; `libaudit.so.*` + `libauparse.so.*` under `/usr/lib`; `/usr/share/audit-rules` stays dropped.

### Runtime evidence (VM)

**auditd service active, enabled by preset, PID 1173:**

auditd service status

**`/var/log/audit/audit.log` written (3864 bytes at capture, root:root 0600). Records include `type=DAEMON_START ver=4.2.1 format=enriched kernel=7.1.8-hadron` and `type=BPF` LOAD/UNLOAD entries:**

audit log directory and first entries

**`auditctl -a always,exit -F arch=b64 -S execve -k probe` accepted; `ausearch -k probe` returns real SYSCALL / EXECVE / PATH records for `tail` and `ausearch`, `arch=c000003e`, `tty=tty1`, `key="probe"`:**

auditctl rule injection and live event

### Config on disk

```
log_file = /var/log/audit/audit.log
log_format = ENRICHED
flush = INCREMENTAL_ASYNC
```

### Musl confirmation

Audit records reference `name="/lib/ld-musl-x86_64.so.1"` — audit-userspace 4.2.1 runs cleanly on Hadron's musl toolchain (upstream `audisp-syslog` needs `#include <unistd.h>` on musl; carried as `patches/0001-audit-syslog-plugin-include-unistd.patch`, submitted upstream at linux-audit/audit-userspace#551).

Screenshots hosted on itxaka-agent/hadron branch `screenshots/audit-597` (fork, not this repo).

mauromorales (approved):

jimmykarily (approved):

tests/image_structure_test.go
```diff
@@ -377,4 +377,87 @@ var _ = Describe("hadron container image structure", Label("image-structure"), f
 		Expect(out).To(MatchRegexp(`(?m)^LOG_OK_LOGINS\s+yes\b`), "LOG_OK_LOGINS should be yes")
 		Expect(out).To(MatchRegexp(`(?m)^UMASK\s+077\b`), "UMASK should be 077")
 	})
+
+	// audit-userspace ships in the full image so kairos-init can drop the
+	// CIS Section 4 baseline rules; see kairos-io/kairos#4637. The stage
+	// builds with `--disable-legacy-actions --without-libcap-ng
+	// --without-io_uring --without-apparmor --without-golang
+	// --without-python3` and installs to /sbin + /etc + /usr/lib.
+	Describe("audit userspace", Label("audit"), func() {
```
itxaka-agent, line 386: None of the six audit specs ever run in CI. The block is labelled `audit` and gated on `skipUnlessFullImage`, but the container job runs `--label-filter "image-structure"` against `hadron-container-amd64` (PR_multiarch.yml:264 and :351), where every one of these specs hits the skip, and the full-image job runs `--label-filter "firmware-layout"` (PR_multiarch.yml:436), which does not select the `audit` label -- so the full image, the only image that actually ships auditd, is never asserted on. Ginkgo's label filter is the DM deciding which checks get rolled: `firmware-layout` gets rolled every build, `audit` sits behind the screen on a sheet nobody looks at.

Wire the label into the full-image job: change the filter on PR_multiarch.yml:436 to "firmware-layout || audit". Alternatively add Label("firmware-layout") to this Describe, but the filter change is clearer since the label list at that step is the one place a reader looks to see what the full image is checked for. Either way, confirm the specs pass against the full image before merging, because right now nothing has executed them.

tests/image_structure_test.go
```diff
@@ -377,4 +377,87 @@ var _ = Describe("hadron container image structure", Label("image-structure"), f
...
+			},
+			Entry("auditctl", "auditctl", "-v"),
+			Entry("ausearch", "ausearch", "--version"),
+			Entry("aureport", "aureport", "--version"),
+			Entry("augenrules", "augenrules", "--version"),
+			Entry("autrace", "autrace", "--version"),
+			// audisp-syslog is the plugin that needed the <unistd.h> patch
+			// (upstream PR linux-audit/audit-userspace#551). If the patch
+			// ever falls out, the binary will not build and this spec fails.
+			Entry("audisp-syslog", "audisp-syslog", "--help"),
```
itxaka-agent, line 420: This table cannot detect a missing binary, so the six entries assert nothing about presence. When the `--entrypoint` binary does not exist, `docker run` exits 127; `runInImage` turns that into `code=127, err=nil` (lines 81-88), and 127 is not in `crashSignalExitCodes`, so the spec passes. Drop `auditctl` from the image entirely and this table stays green. The comment on 404-406 is wrong for the same reason: if the `<unistd.h>` patch falls out, the Docker build fails, not this spec. It is the Death Star sensor sweep reading the Falcon as debris -- exit 127 looks like ordinary junk data, so the table waves it through.

Add an explicit presence spec next to the table, following the precedent already in this file at lines 207-212 ("ships utilities third-party scripts expect"): loop over auditd, auditctl, ausearch, aureport, augenrules, autrace and audisp-syslog with `command -v <bin>` and assert exit 0 for each. Then fix the comment on 404-406 to say the build fails if the patch falls out, since that is the actual guard.

Dockerfile.tmpl
```diff
@@ -2640,6 +2642,48 @@ RUN rm -f \
...
+    --without-golang \
+    --without-libcap-ng \
+    --without-io_uring \
+    --without-apparmor \
+    --with-aarch64 \
+    --with-arm \
+    --with-riscv
+RUN make -s -j${JOBS} -l${MAX_LOAD}
+RUN make -s -j${JOBS} -l${MAX_LOAD} DESTDIR=/audit install
+RUN mkdir -p /audit/etc/audit/rules.d
```
itxaka-agent, line 2782: Builder-confirmed: a real `./configure && make && make install DESTDIR=...` with this PR's exact flags on audit-userspace 4.2.1 produces no `var/` directory at all in DESTDIR -- `/var/log/audit` is never created by `make install` itself. The stage creates `/etc/audit/rules.d` but not `/var/log/audit`, and upstream's auditd.conf points `log_file` at `/var/log/audit/audit.log`. The final stage runs `systemctl preset-all` (Dockerfile.tmpl:4035) and this repo ships no preset policy files, so a shipped auditd.service is enabled by default and starts on first boot; with the log directory absent auditd fails to open its log and the unit dies. auditd arriving on a rootfs with no /var/log/audit is Paul landing on Arrakis without a stillsuit -- the first thing it needs from its environment is the one thing nobody packed.

If `make install` does not already create it, create it here with mode 0700 (that is the permission upstream's packaging uses, and CIS Section 4 expects the directory not to be world-readable). Persistence across A/B upgrades stays with kairos-io/kairos#4629 as the linked issue says; this is only about the directory existing in the image so the unit can start.

```suggestion
RUN mkdir -p /audit/etc/audit/rules.d
RUN install -d -m 0700 /audit/var/log/audit
```

Dockerfile.tmpl
```diff
@@ -2640,6 +2642,48 @@ RUN rm -f \
...
+    --disable-listener \
+    --disable-gssapi-krb5 \
+    --disable-legacy-actions \
+    --without-python3 \
+    --without-golang \
+    --without-libcap-ng \
+    --without-io_uring \
+    --without-apparmor \
+    --with-aarch64 \
+    --with-arm \
```
itxaka-agent, line 2777: `--with-arm` compiles the 32-bit ARM syscall tables into libaudit unconditionally, on every target. Hadron builds three platforms only -- linux/amd64, linux/arm64, linux/riscv64 (REQUIRED_PLATFORMS, populate-sources.yml:41) -- so no image this repo produces can ever run 32-bit ARM binaries. The aarch64 and riscv tables are also baked in on the x86_64 build, which is defensible if it is deliberate (ausearch reading logs collected on another arch) but nothing in the stage says so. It is a chapter keeping a relic Land Raider serviced for a world the Imperium no longer holds: harmless, but someone has to say out loud why it is still in the armoury.

Drop `--with-arm` and keep `--with-aarch64 --with-riscv`, then add a one-line comment above the three flags saying the foreign-arch syscall tables are compiled in on purpose so ausearch/auditctl on one arch can interpret rules and logs from the others. If 32-bit ARM support is wanted for some downstream consumer, say that in the comment instead and keep the flag.

itxaka-agent (changes requested):
> action taken here.

**Verdict: changes-requested.** The Dockerfile plumbing is correct and follows the repo's documented "adding a package" path (`sources.yaml` entry + render substitution + populate-sources gate + sha256 actually enforced by the generated downloader stage), the patch is scoped to exactly the one `#include` its header claims, and the kernel already has `CONFIG_AUDIT`/`CONFIG_AUDITSYSCALL` on all six configs.

The two blockers are both on the test side: the new `Describe` block is labelled `audit`, but CI runs `--label-filter "image-structure"` against the minimal container image (where `skipUnlessFullImage` skips every audit spec) and `--label-filter "firmware-layout"` against the full image — so not one audit assertion has ever executed; and the crash-signal table can't detect a missing binary, so the six new entries prove nothing about presence.

I also routed two build-time claims through the builder before publishing (rule 9b):
- Whether the systemd unit reliably installs: **confirmed** — audit-userspace has no configure-time gating at all, it installs `auditd.service` to `$(prefix)/lib/systemd/system` unconditionally, landing exactly where the new test expects it. That drops the finding I initially drafted about it — the concern doesn't hold up, so it isn't in the list below.
- Whether `make install` creates `/var/log/audit`: **contradicted** — a real build shows `make install` creates no `var/` directory at all. That confirms the directory really is missing from the image unless the Dockerfile creates it explicitly, so that finding stays, below, with the evidence folded in.

Four findings, inline on the diff.

itxaka-agent:
> action taken here.

**Audit trail — round 0**

1. **Intake.** Picked up this PR for review (unassigned, no reviews yet). Pulled `add_audit` into a scratch worktree at `HEAD 209c7e9`.
2. **Reviewer, round 0.** Verdict: `changes-requested`. Confirmed the Dockerfile plumbing follows the repo's existing package-add convention, the out-of-tree patch is scoped to exactly the one `#include` it claims, and kernel audit support is already enabled on all six configs. Flagged five potential issues, two of which needed empirical build verification before publishing (rule 9b).
3. **Builder verification, finding "systemd unit install."** Claim tested: does `make install` create `/usr/lib/systemd/system/auditd.service` under this PR's exact configure flags? **Confirmed** — audit-userspace installs the unit unconditionally, no configure-time gating exists. That result disproves the finding's own premise (that gating might skip the unit), so the finding was dropped rather than published — the test in question will actually pass.
4. **Builder verification, finding "/var/log/audit directory."** Claim tested: does `make install` create `/var/log/audit`? **Contradicted** — a real build shows `make install` creates no `var/` directory at all. That confirms the underlying concern is real: without an explicit `mkdir`, the directory genuinely will not exist in the image. Finding published with the evidence folded in.
5. **Review posted.** [Round 0 review](https://github.com/kairos-io/hadron/pull/597#pullrequestreview-5197821640), `CHANGES_REQUESTED`, 4 inline comments on the diff:
   - `tests/image_structure_test.go:386` — the new `audit` label is never selected by either CI job's `--label-filter`, so none of the six new specs have ever run.
   - `tests/image_structure_test.go:407` — the crash-signal table can't detect a missing binary (a `127` exit from a nonexistent entrypoint isn't a tracked crash signal), so the table doesn't actually prove the binaries are present.
   - `Dockerfile.tmpl:2683` — `/var/log/audit` isn't created anywhere, and `auditd.conf` needs it; a default-enabled unit will fail to start without it. Suggested a one-line `install -d -m 0700` fix.
   - `Dockerfile.tmpl:2679` — `--with-arm` bakes in 32-bit ARM syscall tables on a build that only ever targets amd64/arm64/riscv64; flagged as likely dead weight unless there's a reason to keep it.

This PR stays in `awaiting-author`. Next slot re-polls: if the branch gets a new commit, review resumes at round 1; if nothing changes for 7 days, this escalates per rule 18.

Dockerfile.tmpl
Itxaka: this is kind of true

itxaka-agent:
@Itxaka fixes for the 4 review findings + the missing MAX_LOAD scope fix are on `itxaka-agent/hadron:add_audit` (cde4a1c14af69cc4188724a1dcd7d13473e66089). Cannot push directly to `kairos-io/hadron` (403 on force-push). To land, run:

```
git fetch https://github.com/itxaka-agent/hadron add_audit
git checkout add_audit
git reset --hard FETCH_HEAD
git push --force-with-lease origin add_audit
```

What the commit does (`fix(audit): address review findings and pick up MAX_LOAD scope fix`):

- Rebases onto main to pull in [ba6aadc](https://github.com/kairos-io/hadron/commit/ba6aadc) — the MAX_LOAD scoping fix. Audit stage now uses `${MAX_LOAD:+-l${MAX_LOAD}}` and declares `ARG MAX_LOAD`, so `make test-render` is green (verified locally: `274 stages checked in Dockerfile; MAX_LOAD in scope wherever it is used, every use guarded, no bare -l`).
- Drops `--with-arm` from the configure line — Hadron builds no 32-bit ARM per `REQUIRED_PLATFORMS`. Adds an in-line comment saying the aarch64/riscv foreign-arch tables are compiled in on purpose so ausearch/auditctl on one arch can read logs from another.
- Wires `audit` into the full-image ginkgo label filter in `PR_multiarch.yml:454` — changes `"firmware-layout"` to `"firmware-layout || audit"` so the specs actually run against the image that ships auditd.
- Adds an explicit `command -v` presence spec covering `auditd`, `auditctl`, `ausearch`, `aureport`, `augenrules`, `autrace`, `audisp-syslog` in `tests/image_structure_test.go`. Same shape as `ships utilities third-party scripts expect` at lines 207-212. The `DescribeTable` stays for crash detection; the new `It` catches missing binaries the crash table misses (exit 127 is not a crash signal).
- Fixes the audisp-syslog comment to say the Docker build fails if the `<unistd.h>` patch falls out, not the spec.

`/var/log/audit` directory create (`install -d -m 0700`) was already on b27cdd1 and is preserved by the rebase.

---

*Automated triage-agent (`kairos-triage-agent`) note. Cannot self-merge into `kairos-io/hadron`; needs your push.*

Dockerfile.tmpl
```diff
@@ -2735,6 +2737,53 @@ RUN rm -f \
...
+    --without-python3 \
+    --without-golang \
+    --without-libcap-ng \
+    --without-io_uring \
+    --without-apparmor \
+    ## Foreign-arch syscall tables compiled in on purpose so ausearch and
+    ## auditctl on one arch can read rules and logs collected on another.
+    ## Hadron builds no 32-bit ARM (REQUIRED_PLATFORMS: amd64/arm64/riscv64),
+    ## so --with-arm is omitted.
+    --with-aarch64 \
```
itxaka-agent, line 2778: Builder-confirmed (v4.2.1 build+run, `uname()` faked to aarch64): rejecting `-F arch=b32` without `--with-arm` surfaces as "elf mapping not found" (`EAU_ELFUNKNOWN`) via libaudit's `audit_rule_fieldpair_data()`, while the identical build with `--with-arm` accepts it. Dropping `--with-arm` removes the ARM entry from libaudit's machine table, and on aarch64 `-F arch=b32` resolves to `AUDIT_ARCH_ARM`, so `auditctl`/`augenrules` reject any rule carrying `arch=b32` on an arm64 image. The CIS Section 4 rule text is written as `arch=b64`/`arch=b32` pairs, so when kairos-io/kairos#4626 drops those files into `/etc/audit/rules.d` the arm64 image loads half of each pair and reports the other half as a bad rule; this happens at rule-parse time in userspace, so it does not depend on the arm64 kernel having `CONFIG_COMPAT` or on anything 32-bit ever executing. This corrects my own round-0 advice, not a new requirement I'm inventing: I offered keeping the flag as the alternative, and this is the downstream consumer that wants it. Warhammer 40k: we scrapped the relic Land Raider I called useless, and then the Codex's standing battle order turned out to list it in every formation.

Put `--with-arm \` back in the flag list above `--with-aarch64`, and rewrite the comment on 2774-2777 so it says the 32-bit ARM table is kept because CIS Section 4 rules are b64/b32 pairs and `arch=b32` on aarch64 means `AUDIT_ARCH_ARM`. If you'd rather not carry the table, say in the comment that arm64 images can't load `arch=b32` rules and raise it on kairos-io/kairos#4626 so the rule files get an arch-conditional variant instead of failing to load.

itxaka-agent (changes requested):
> action taken here.

**Verdict: changes-requested (round 1).** All four round-0 findings are closed: the `audit` label now runs in the full-image job (`PR_multiarch.yml:454`), the new `command -v` presence spec catches a missing binary the crash-signal table couldn't, the `--with-arm` flag change is explained, and `audisp-syslog`'s comment is accurate. `/var/log/audit` (mode 0700) landed correctly from the earlier suggestion.

One new blocker, and it's a correction of my own round-0 suggestion: dropping `--with-arm` breaks `arch=b32` audit rules on arm64 (see inline comment). I routed the claim through the builder before publishing (rule 9b) rather than publish it on source-reading alone, since the original `--with-arm` finding was mine to begin with. A second draft finding, about the `##` comment lines mid-continuation possibly swallowing the `--with-aarch64`/`--with-riscv` flags, was builder-tested and contradicted — a real `docker build`/`podman build` reproduction shows the flags reach `configure` intact — so it isn't in the list below.

One finding, inline on the diff.

itxaka-agent:
> action taken here.

**Round-1 audit summary.**

- **Trigger:** you pushed `cde4a1c` addressing the four round-0 findings (CI label-filter gap, missing binary-presence spec, unexplained `--with-arm`, stale `audisp-syslog` comment); `/var/log/audit` had already landed in `f7a515c` from the earlier suggestion block.
- **Reviewer (round 1):** verified each of the four fixes against the new diff rather than trusting the commit message. All four hold up.
- **New finding:** removing `--with-arm` (my own round-0 suggestion) breaks `arch=b32` audit rules on arm64. Routed through the builder before publishing since I raised the original suggestion myself — confirmed by a real build+run (rejects `-F arch=b32` with `EAU_ELFUNKNOWN` on aarch64 without the flag, accepts it with the flag). Posted inline on `Dockerfile.tmpl:2778` with the evidence.
- **Dropped finding:** a second draft finding worried the `##` comment lines mid-continuation in the `./configure` block might swallow `--with-aarch64`/`--with-riscv`. Builder-tested with a real `docker build`/`podman build` reproduction — the flags reach `configure` intact, no bug. Not published.
- **Verdict:** changes-requested, one inline comment.

Review: https://github.com/kairos-io/hadron/pull/597#pullrequestreview-5198440798

Dockerfile.tmpl
Itxaka: @copilot can you fix this

Dockerfile.tmpl
```diff
@@ -2735,6 +2737,53 @@ RUN rm -f \
...
+    --disable-legacy-actions \
+    --without-python3 \
+    --without-golang \
+    --without-libcap-ng \
+    --without-io_uring \
+    --without-apparmor \
+    ## Keep 32-bit ARM table enabled: CIS Section 4 rules are shipped as
+    ## arch=b64/arch=b32 pairs, and on aarch64 arch=b32 resolves to
+    ## AUDIT_ARCH_ARM at userspace rule-parse time.
+    --with-arm \
```
itxaka-agent, line 2777: The flag change on this line is correct and resolves the round-1 finding; the problem is the commit that carries it. Commit 3920347 ("fix(audit): restore --with-arm for aarch64 b32 rule parsing", pushed by copilot-swe-agent[bot]) has no `Signed-off-by` trailer -- its only trailer is `Co-authored-by: Itxaka <1447686+Itxaka@users.noreply.github.com>`. The other three commits on the branch (e4148f2, f7a515c, cde4a1c) each carry a `Signed-off-by`, so this one commit is the DCO gap in an otherwise signed-off branch, and no developer has certified it. Mass Effect: the fix itself is a clean Normandy refit, but the requisition order came through unsigned -- Hackett's office does not care how good the upgrade is if nobody put their name on the paperwork.

Amend 3920347 to add the sign-off and force-push: `git commit --amend --no-edit -s` (or `git -c user.name=... commit --amend --no-edit --trailer 'Signed-off-by: Itxaka <itxaka@kairos.io>'`), then `git push --force-with-lease`. Keep the existing `Co-authored-by: Itxaka` line and the subject as-is -- both are fine. Do not change any code: --with-arm at line 2777 and the comment at 2774-2776 are what I asked for in round 1 and they are correct.

itxaka-agent (changes requested):
> This comment was generated by software. A human maintainer can override
> any action taken here.

Round 2: `3920347` correctly restores `--with-arm` and fixes the `arch=b32` regression I raised in round 1 — I re-read the diff and the new comment text, and it matches the builder-confirmed mechanism. The only remaining issue is that this commit carries no `Signed-off-by` trailer, unlike the other three commits on the branch.

itxaka-agent:
> This comment was generated by software. A human maintainer can override
> any action taken here.

**Audit trail — round 2**

1. **Trigger.** After the round-1 review, you replied on the `--with-arm` finding asking `@copilot` to fix it. `copilot-swe-agent[bot]` (co-authored by you) pushed `3920347`, "fix(audit): restore --with-arm for aarch64 b32 rule parsing".
2. **Reviewer, round 2.** Read the new diff directly rather than trusting the commit title: `--with-arm` is restored at `Dockerfile.tmpl:2777`, and the rewritten comment above it correctly states the CIS Section 4 `arch=b64`/`arch=b32` pair mechanism — matches the round-1 builder-confirmed evidence. Re-verified the four round-0 fixes are still intact (diff-stat arithmetic checks out: no unrelated deletions).
3. **New finding.** Commit `3920347` carries no `Signed-off-by` trailer — only `Co-authored-by: Itxaka`. The other three commits on the branch (`e4148f2`, `f7a515c`, `cde4a1c`) all have one. This is a DCO gap, not a code issue.
4. **Review posted.** [Round 2 review](https://github.com/kairos-io/hadron/pull/597#pullrequestreview-5206237494), `CHANGES_REQUESTED`, 1 inline comment on `Dockerfile.tmpl:2777` asking for the sign-off trailer to be amended onto that commit.

This PR stays in `awaiting-author`. Next slot re-polls: if the branch gets a new commit, review resumes at round 3 (the last round before the cap); if nothing changes for 7 days, this escalates per rule 18.

Dockerfile.tmpl
```diff
@@ -2735,6 +2737,54 @@ RUN rm -f \
...
+    --with-aarch64 \
+    --with-riscv
+RUN make -s -j${JOBS} ${MAX_LOAD:+-l${MAX_LOAD}}
+RUN make -s -j${JOBS} ${MAX_LOAD:+-l${MAX_LOAD}} DESTDIR=/audit install
+RUN mkdir -p /audit/etc/audit/rules.d
+RUN install -d -m 0700 /audit/var/log/audit
+## Drop upstream example rule templates and audit aclocal macros.
+RUN rm -rf /audit/usr/share/audit-rules \
+ && rm -f /audit/usr/share/aclocal/*audit*.m4
```
itxaka-agent, line 2786: Five commits on this branch carry no `Signed-off-by` trailer: 3920347, 58398f7, d79ca9c (which adds this line), 6db0595 and 607864a. Their only trailer is `Co-authored-by: Itxaka <1447686+Itxaka@users.noreply.github.com>`. The first three commits (e4148f2, f7a515c, cde4a1c) each carry a sign-off, so every commit pushed by copilot-swe-agent[bot] since round 2 is uncertified, and the count has gone from one to five. The DCO check reporting "All commits are signed off!" is not evidence to the contrary: the trailer is absent from the commit objects, so whatever that check matches on, it is not the sign-off line, and this repo carries no `.github/dco.yml` or DCO workflow that would explain an exemption. Warhammer 40k: the Administratum stamped the requisition green without reading it, and five munitions shipments left the forge world with the certification field blank.

Rebase the five and add the trailer: `git rebase --exec 'git commit --amend --no-edit -s' cde4a1c` then `git push --force-with-lease`. While the branch is being rewritten, fold 58398f7, d79ca9c, 6db0595 and 607864a into the commits they fix — they are CI fixups for this PR's own new code, and `chore: finalize ci failure fix validation` names a process step rather than a change, so it should not survive as a subject on main. If you would rather not rewrite, squash-merge instead and put `Signed-off-by: Itxaka <itxaka@kairos.io>` in the squash body; the unsigned intermediates then never land on main. Keep the `Co-authored-by` lines either way.

tests/image_structure_test.go
```diff
@@ -377,4 +377,100 @@ var _ = Describe("hadron container image structure", Label("image-structure"), f
...
+			Expect(out).To(ContainSubstring("libaudit.so."))
+			Expect(out).To(ContainSubstring("libauparse.so."))
+		})
+
+		It("does NOT ship dropped upstream artifacts", func() {
+			// The audit stage removes rule templates and its aclocal macros
+			// to keep the image small (see Dockerfile audit stage). Guard
+			// the deletion so a bump does not accidentally reintroduce them.
+			out, code := shInImage(
+				"! test -e /usr/share/audit-rules && ( ! test -d /usr/share/aclocal || ! find /usr/share/aclocal -maxdepth 1 -name '*audit*.m4' | grep -q . ) && echo OK")
```
itxaka-agent, line 470: The aclocal half of this assertion cannot fail, so the two commits that produced it (d79ca9c narrowing the deletion, 6db0595 loosening the check) guard nothing. `full-image-merge-base` copies the audit payload at Dockerfile.tmpl:3986 and then runs `find /skeleton -type f -name '*.m4' -delete` at Dockerfile.tmpl:3998 — same stage, sweep after the copy, no COPY in between — so no `.m4` file of any name reaches the full image whatever the audit stage removed. Delete `rm -f /audit/usr/share/aclocal/*audit*.m4` at Dockerfile.tmpl:2786 and this spec still passes. D&D: a Perception check at DC 0 — line 3998 already cleared the room, so the roll succeeds no matter what the audit stage did.

Drop the aclocal clause and keep the spec to the one artifact the audit stage owns. Then delete `rm -f /audit/usr/share/aclocal/*audit*.m4` at Dockerfile.tmpl:2786 (leaving `RUN rm -rf /audit/usr/share/audit-rules` with no line continuation) and reword the comment at 2784 to mention only the rule templates, since the global `*.m4` sweep at 3998 already covers the macros.

```suggestion
				"! test -e /usr/share/audit-rules && echo OK")
```

itxaka-agent (changes requested):
> action taken here.

Round 3. Four new commits (58398f7, d79ca9c, 6db0595, 607864a) landed on top of 3920347 since the round-2 review, apparently to fix a CI failure. The round-2 finding about missing sign-off is unresolved and has grown: none of the four new commits carry a `Signed-off-by` trailer either, so the branch now has five unsigned commits instead of one. One of the four new commits also introduces a redundant test assertion that a later Dockerfile stage already makes unconditionally true. A third possible finding (whether `rm -rf /audit/usr/share/audit-rules` targets a real path) was checked empirically against a real audit-userspace 4.2.1 build with this PR's exact configure flags — it does, so that one is not included below.

itxaka-agent:
> action taken here.

**Round-3 audit summary.**

- **Trigger.** Since the round-2 review, copilot-swe-agent[bot] pushed four new commits (`58398f7`, `d79ca9c`, `6db0595`, `607864a`) on top of `3920347`, apparently to fix a CI failure (test/build alignment for the audit-userspace image-structure checks). None address the round-2 finding.
- **Round-2 finding still open, and worse.** `3920347` still has no `Signed-off-by` trailer, and none of the four new commits do either — the branch now carries 5 unsigned commits instead of 1, all with only `Co-authored-by: Itxaka <...>`. GitHub's DCO check on this PR reports success, but the raw commit objects (checked with `git cat-file -p`) confirm the trailer is genuinely absent; suggested a rebase with `git commit --amend --no-edit -s` (and folding the CI-fixup commits into what they fix) or a squash-merge with the trailer in the squash body.
- **New finding.** `tests/image_structure_test.go:470`'s aclocal half is redundant — a later Dockerfile stage (`Dockerfile.tmpl:3998`) already deletes every `.m4` file from the full image regardless of what the audit stage does, so that half of the assertion can never fail. Posted with a suggestion block to drop it.
- **One finding checked and dropped.** Whether `rm -rf /audit/usr/share/audit-rules` (Dockerfile.tmpl:2785) targets a real path was verified against an actual audit-userspace 4.2.1 build with this PR's exact configure flags: it does — `rules/Makefile.am` hardcodes that as the install location, and no `/usr/share/audit/sample-rules` path exists at all. Not published.
- **Verdict: changes-requested**, 2 of 3 draft findings posted inline per rule 12a (`Dockerfile.tmpl:2786`, `tests/image_structure_test.go:470` with a suggestion block).

Review: https://github.com/kairos-io/hadron/pull/597#pullrequestreview-5210236631

PR #599, opened 2026-09-14
ci: split buildx cache per stage on amd64 and arm64

itxaka-agent:
PR builds barely hit cache even for trivial changes. On PR #594 (a
react bump that touches nothing the container build sees) toolchain-amd64
came back with 182 of 924 steps cached (~16%), container-amd64 with
570/796 (~42%), hadron-bios-amd64 cloud/no-fips with 399/1245 (~24%).
On that PR we should be seeing ~99%.

The cause is that amd64 and arm64 only wrote 3-4 monolithic quay tags
per arch — `toolchain-<arch>`, `container-<arch>`,
`bios-<kernel>-<fips>-<arch>`, `trusted-<kernel>-<arch>` — each
`mode=max` for its `--target` and packing 100+ intermediate stages
under a single OCI manifest. At the end of a ~60 minute main build
the whole thing has to push in one shot, and we deliberately keep
`ignore-error=true` on cache-to so a quay flake doesn't red main. The
side effect is that a truncated or failed push is swallowed, the next
PR imports a partial manifest, and buildkit re-runs everything past
whatever survived — usually from mid-toolchain.

This mirrors the per-stage cache layout that PR_riscv64.yml and the
riscv64 half of build-multiarch-images.yml already use. New jobs per
arch (amd64 and arm64): stage0, stage1 (`stage1-merge`), rsync, cmake,
kernel (matrix over cloud and default, `kernel-misc`), systemd and
python (`python-build`). Each writes its own small tag. The downstream
jobs (toolchain, container, hadron-bios, hadron-trusted,
hadron-grub-arm64) grew a cache-from list that pulls every upstream
stage tag before falling back to their own aggregate tag.

The PR pipeline is the same shape with `pr-<num>-<stage>-<arch>` tags
gated on `PR_CACHE_WRITE`. pr-cache-cleanup.yml already matches
`pr-<num>-*` so nothing there needs changing. `ignore-error=true`
stays on cache-to.

Runner cost: main peak per arch goes from 3 to 7 concurrent
oracle-vm-32cpu-128gb runners, same on PR. riscv64 is untouched
because it already looks like this.


itxaka-agent:
> This comment was generated by software. A human maintainer can override any action taken here.

@Itxaka fix for [`ci-robbot`'s note](https://github.com/kairos-io/hadron/pull/599#issuecomment-5687147756) is on `itxaka-agent/hadron:split-cache-per-stage-amd64-arm64` (a3d78bfa2951c7350041cb0d3f0c02421f729c5e). Cannot force-push to `kairos-io/hadron` (fork-only access). To land, from a checkout with push:

```
git fetch https://github.com/itxaka-agent/hadron split-cache-per-stage-amd64-arm64
git checkout split-cache-per-stage-amd64-arm64
git reset --hard FETCH_HEAD
git push --force-with-lease https://github.com/kairos-io/hadron split-cache-per-stage-amd64-arm64
```

Commit `fix(cache): resolve matrix.kernel via format(), drop dead kernel-misc tag`:

- `.github/workflows/build-multiarch-images.yml:293` (kernel-amd64) and `:672` (kernel-arm64): `${{ matrix.kernel }}` was nested inside a `format()` argument, so the tag string never expanded and both jobs pushed a cache tag with the literal `${{ matrix.kernel }}` baked in. Rewrote both to thread `matrix.kernel` as a positional argument (`format('...kernel-misc-{1}-<arch>,{0}', env.CACHE_TO, matrix.kernel)`) — the same shape every other cache-to in this file and `PR_multiarch.yml` already uses.
- Dropped the unversioned `kernel-misc-<arch>` entry from `cache-from` in both jobs. No job writes that tag, so every kernel-misc build paid for one extra pull that always missed.

---

*Automated triage-agent note. Cannot self-merge into `kairos-io/hadron`; needs your push.*

itxaka-agent:
Perms updated — pushed `a3d78bf` directly to `kairos-io/hadron:split-cache-per-stage-amd64-arm64` (`452a9b3..a3d78bf`). Fork-handoff detour no longer needed. CI will re-run against the new head.

---

*Automated triage-agent (`kairos-triage-agent`) status update.*

Itxaka (approved):

PR #593, opened 2026-09-09
fix(ci): fail a congested setup-go cache restore fast

ci-robbot:
Fixes kairos-io/kairos#4564.

Five of the last 40 `PR_multiarch` runs lost a job to `Install Go` before a
single test ran, always after burning exactly 5m00s:

| run | job |
| --- | --- |
| [34314852826](https://github.com/kairos-io/hadron/actions/runs/34314852826) | reset tests (FIPS) |
| [34294143535](https://github.com/kairos-io/hadron/actions/runs/34294143535) | reset tests (FIPS) |
| [34162208109](https://github.com/kairos-io/hadron/actions/runs/34162208109) | acceptance tests |
| [34141532218](https://github.com/kairos-io/hadron/actions/runs/34141532218) | acceptance tests (FIPS) |
| [33888029300](https://github.com/kairos-io/hadron/actions/runs/33888029300) | reset tests (FIPS) |

Exactly 5m00s is the step's own `timeout-minutes: 5`, so this is not the
toolchain download. That comes from the runner tool cache in seconds. The
rest of the step is `setup-go` restoring a ~305 MB dependency cache, and on
run 34314852826 it was at 87.8% moving at 0.9 MB/s when the cap fired.

## The change

Per @Itxaka's review: the workers cap at 100 Mb/s and a congested one is not
going to get through the image pulls later either, so the right move is to
give up early rather than to wait longer.

I sampled the step across the last 12 `PR_multiarch` runs, 92 executions of
`Install Go` in total, and the split is clean:

| | duration |
| --- | --- |
| median (91 successes) | 17s |
| p90 | 41s |
| slowest healthy | 52s |
| congested outliers | 277s (passed), 300s (timed out) |

Nothing healthy comes near a minute. So the six active `setup-go` steps drop
from `timeout-minutes: 5` to `2`, with the numbers recorded at each so the
value is not re-guessed later.

At two minutes a healthy restore keeps 2.3x headroom over the slowest one
observed, and a doomed job gives back three minutes of runner time. The cost
is the 277s case: that job did eventually pass, and now it would not. That
looks like the right trade, since it was already 5x slower than any healthy
run and the same congestion had four other jobs failing outright.

Verified the six edits all land inside `actions/setup-go` steps by parsing
the workflow and walking `jobs[*].steps[*]` (`container-amd64`,
`container-arm64`, `hadron-bios-amd64`, `tests-bios-amd64`,
`tests-trusted-amd64`, `tests-bios-fips-amd64`), which covers every job
[...]

PR #600, opened 2026-09-16
fix(ci): install Go without the cache instead of failing the leg

### LINKED ISSUE #4564: hadron CI: setup-go cache restore times out and reddens 12.5% of PR_multiarch runs
## The bug

12.5% of `PR_multiarch` runs (5 of the last 40) lose a job to `Install Go`,
before a single test executes. It is not the toolchain download. It is
`actions/setup-go` restoring its dependency cache and running out the step's
`timeout-minutes: 5`.

The signature is unmistakable: the step burns exactly 5m00s, every time.

| run | job | Install Go |
| --- | --- | --- |
| [34314852826](https://github.com/kairos-io/hadron/actions/runs/34314852826) | reset tests (FIPS) | 07:45:20 -> 07:50:20 |
| [34294143535](https://github.com/kairos-io/hadron/actions/runs/34294143535) | reset tests (FIPS) | 5m00s |
| [34162208109](https://github.com/kairos-io/hadron/actions/runs/34162208109) | acceptance tests | 22:41:42 -> 22:46:42 |
| [34141532218](https://github.com/kairos-io/hadron/actions/runs/34141532218) | acceptance tests (FIPS) | 17:35:55 -> 17:40:55 |
| [33888029300](https://github.com/kairos-io/hadron/actions/runs/33888029300) | reset tests (FIPS) | 16:44:38 -> 16:49:38 |

On run 34314852826 the cache restore was at 87.8% of 305 MB when the cap
fired, moving at 0.9 MB/s. At that rate the restore needs about 5m40s, so it
loses to a 5 minute cap by a whisker. Nine jobs in the workflow use
`setup-go`, and any of them can draw the short straw.

Because these are `needs:`-ed test jobs, one timeout turns the whole run red
on a PR whose diff has nothing to do with it. That is what happened twice in
a row on kairos-io/hadron#592.

## The fix

kairos-io/hadron#593 raises the cap on all six active `setup-go` steps from 5
to 15 minutes. `timeout-minutes` only ever aborts a step, so this costs a
healthy run nothing: when the runner network is fine the same step finishes
in seconds and never approaches either number. It only stops a slow-but-
working restore from being killed.

## The part I am deliberately leaving alone

The deeper question is whether a 305 MB cache earns its keep here at all.
`tests/go.mod` is a small Ginkgo suite, 95 modules in `go.sum`, so the bulk
of that 305 MB is `~/.cache/go-build`, not the module set. The cache is a
clear win when the network is healthy and actively worse than no cache when
it is not.

Caching only `~/go/pkg/mod` and dropping the build cache would cut the
restore to a fraction of the size, at the cost of recompiling the test binary
in every job. That is a performance trade I cannot measure from here, and it
wants someone who can benchmark a full run on the ARC runners. Same shape as
#4559. Happy to open that PR if someone picks a direction.

ci-robbot:
Follow-up to #593, which closed kairos-io/kairos#4564.

#593 capped the `setup-go` dependency-cache restore at two minutes, on the
reasoning that a worker too congested to pull 305MB in a minute would not
manage the image pulls later either. The first half held; the second did not.

Run [35105248251](https://github.com/kairos-io/hadron/actions/runs/35105248251)
lost `recovery tests` and `acceptance tests` to
`The action 'Install Go' has timed out after 2 minutes`, each preceded by
`Failed to restore: The operation cannot be completed in timeout`. Matrix
fail-fast then cancelled `reset tests` and `custom-partitioning tests`. Four
of twelve legs, and a re-run of a ~2h matrix, on a cache restore.

**The same run is the argument.** Every `Install Go` step in it, same cache
key, same commit:

| leg | Install Go |
| --- | --- |
| `container-arm64` | 8s |
| `container-amd64` | 9s |
| `hadron-bios-amd64` x3 | 13-14s |
| `custom-partitioning tests` | 10s |
| `reset tests` | 37s |
| `acceptance tests (FIPS)` | 40s |
| `recovery tests` | **120s, capped** |
| `acceptance tests` | **120s, capped** |

The four healthy runs since #593 merged are all 0-14s across every leg. So the
cap is measuring the worker it landed on, and the leg should degrade instead of
failing.

## The change

`continue-on-error: true` on the capped step, plus a second `setup-go` with
`cache: false` gated on `steps.install-go.outcome == 'failure'`. Six steps, the
six that #593 capped.

A congested worker then costs that job a module download from the proxy rather
than a red check. The second step repeats the toolchain half, which is a
`hostedtoolcache` lookup and costs seconds, and skips the 305MB that was the
problem.

The two-minute cap stays, per @Itxaka's direction on #4564: nothing here waits
longer than before.

`continue-on-error` hides nothing. A genuinely broken `go-version-file` or
unavailable toolchain fails the second step, which has no
`continue-on-error`, so the leg still reds for a real cause.

## Verified

The workflow parses, and each of the six pairs was asserted programmatically
rather than by eye: primary step has `id: install-go`,
`continue-on-error: true` and `timeout-minutes: 2`; the step immediately after
it is the fallback, with the right `if:`, `cache: false` and
`go-version-file`. `git diff` hunks are confined to the six blocks, and the
commented-out `tests-bios-arm64` job is untouched.

Behaviour under a cap hit cannot be reproduced on demand; it needs a congested
ARC worker. What CI proves here is the healthy path, where the fallback step is
skipped and nothing changes.

The remaining question from #4564 is still open and not addressed here: most of
the 305MB is `~/.cache/go-build` rather than `tests/go.mod`'s 95 modules, so
caching only `~/go/pkg/mod` would shrink the restore at the cost of a recompile
per job. That needs a benchmark on these runners.


Itxaka:
no this is useless. When the cache fails its usually because the node is network congested, this only moves the issue into later. Installing the go packages still needs to pull them and their deps, so we would be moving the slowness into a newer step. 

Anyway, this may be fixed if the workers that we use are not ours anymore and we are using the cncf ones, as this usually only happened on our kairos runnners,

PR #608, opened 2026-09-17
Automatic version bumps for expat, libelf, libxml2, pcre2

ci-robbot:
## Summary

- expat was updated 2.8.2 -> 2.8.4
- libelf was updated 0.195 -> 0.196
- libxml2 was updated 2.15.3 -> 2.15.4
- pcre2 was updated 10.47 -> 10.48

### expat
- Version: `2.8.2` -> `2.8.4`
- Release: [libexpat/libexpat 2.8.4](https://github.com/libexpat/libexpat/releases/tag/R_2_8_4)
- Notes:

<details>
<summary>Release notes excerpt</summary>

* [Blog post "Expat 2.8.4 released, fixes 4 vulnerabilities"](https://blog.hartwork.org/posts/expat-2-8-4-released/)
* [Original change log](https://github.com/libexpat/libexpat/blob/R_2_8_4/expat/Changes)

</details>

- Compare: [R_2_8_2...R_2_8_4](https://github.com/libexpat/libexpat/compare/R_2_8_2...R_2_8_4)

### libelf
- Version: `0.195` -> `0.196`
- Release: [arachsys/libelf releases](https://github.com/arachsys/libelf/releases)
- Compare: [v0.195...v0.196](https://github.com/arachsys/libelf/compare/v0.195...v0.196)

### libxml2
- Version: `2.15.3` -> `2.15.4`
- Notes:

<details>
<summary>Release notes excerpt</summary>

libxml 2.15.4

</details>

- Compare: [v2.15.3...v2.15.4](https://gitlab.gnome.org/GNOME/libxml2/-/compare/v2.15.3...v2.15.4)

### pcre2
- Version: `10.47` -> `10.48`
- Release: [PCRE2Project/pcre2 10.48](https://github.com/PCRE2Project/pcre2/releases/tag/pcre2-10.48)
- Notes:

<details>
<summary>Release notes excerpt (truncated)</summary>

This is a regular release, incorporating security fixes along with fixes and small improvements to library behaviour.

This release is available as before as a (signed) Git tag, or alternatively as a (signed) tarball of the Git tag ([attestation](https://github.com/PCRE2Project/pcre2/attestations/44153304)).

### NEWS

Only changes to behaviour, changes to the API, and other significant changes are described here. Please see the [ChangeLog](https://github.com/PCRE2Project/pcre2/blob/pcre2-10.48/ChangeLog) and Git log for further details.

* (Git change) Renamed the default development branch from master to main.

[...]

PR #601, opened 2026-09-17
Automatic version bumps for libtirpc, multipath, parted, urcu

ci-robbot:
## Summary

- libtirpc was updated 1.3.7 -> 1.3.8
- multipath was updated 0.15.0 -> 0.15.1
- parted was updated 3.7 -> 3.7.13
- urcu was updated 0.15.6 -> 0.15.7

### libtirpc
- Version: `1.3.7` -> `1.3.8`
- Source: [upstream](git://git.linux-nfs.org/projects/steved/libtirpc.git) (gittag)
- Notes: unavailable for this source type

### multipath
- Version: `0.15.0` -> `0.15.1`
- Release: [opensvc/multipath-tools 0.15.1](https://github.com/opensvc/multipath-tools/releases/tag/0.15.1)
- Notes:

<details>
<summary>Release notes excerpt (truncated)</summary>

### User-visible changes

* The `preferredip=`parameter for the `iet` prioritizer has been generalized.
  The new syntax is
  ```
  preferredip=<CIDR>:<Prio>[,<CIDR>:<Prio>,...]
  ```
  where `CIDR` is an IPv4 address block in CIDR format, e.g. `192.168.1.0/24`,
  and `<Prio>` is the numeric priority to assign to IP addresses in this
  block.  See the *multipath.conf(5)* man page for details. The syntax
  `preferredip=<IP>` is still supported with the same semantics as before.

### Vulnerability fixes

* [DoS on mulipathd socket by blocking IPC send operations](https://github.com/opensvc/multipath-tools/security/advisories/GHSA-hmcm-9cq4-r2xm)
* [DoS on multipathd socket by exhausting connections]( https://github.com/opensvc/multipath-tools/security/advisories/GHSA-pvp6-c9p3-25fp)
* [Heap Out-of-Bounds Read in Custom Format String Parser via Trailing `%`](https://github.com/opensvc/multipath-tools/security/advisories/GHSA-g5mh-253r-jjw5)
* [Heap out-of-bounds read in device-mapper-multipath ALUA RTPG parsing](https://github.com/opensvc/multipath-tools/security/advisories/GHSA-pxwh-g75c-95pc)
* [kpartx: Heap Out-of-Bounds Read in GPT Header Validation](https://github.com/opensvc/mu…

</details>

- Compare: [0.15.0...0.15.1](https://github.com/opensvc/multipath-tools/compare/0.15.0...0.15.1)

### parted
- Version: `3.7` -> `3.7.13`
[...]

PR #609, opened 2026-09-17
fix(autobumper): refresh the sources.yaml checksum a version bump invalidates

ci-robbot:
## The bug

`849b115` made the Dockerfile the single source of truth for pinned versions, and
its own message describes a bump as "one ARG default plus one sha256". Only the
first half exists. Every `updatecli.d/*.yaml` target is `kind: dockerfile` on an
`ARG <NAME>_VERSION` matcher, and nothing in the tree writes `sources.yaml`.

So a bump leaves the checksum describing the previous release, and both readers
of that field reject the new tarball:

- `populate-sources` fails the cell on a checksum mismatch.
- The downloader stages `hack/render.sh` writes for fork pull requests exhaust
  their URL list without a match, which is how the nightly's own PRs build.

That is why #601 through #608 are red, each on its own bumped package's source
download step. Verified directly:

```
sources.yaml pins: c08ae2388ef333e8403e670ad70c0a11f1eed021fd88308d7e02f596fcd9dc16
pcre2 10.47:       c08ae2388ef333e8403e670ad70c0a11f1eed021fd88308d7e02f596fcd9dc16
pcre2 10.48:       ebcc25aadf2a51fa1fefa9b8bc9e7a79b3dae86870a0f1152a22e42befd46888   <- what #608 pins
```

## The fix

`hack/refresh-source-checksums.sh` resolves each package's version from the
Dockerfile ARG default (same rule `render.sh`, `list-missing-sources.sh` and
`populate-sources` already use), downloads the tarball from the first URL that
serves it, and rewrites only that package's `sha256:` line. The autobumper runs
it with `--changed` right after `updatecli apply`, before
`make update-kernel-configs` so a kernel version no URL serves is reverted
before the configs are regenerated against it.

A package whose new version no URL serves keeps its checksum and has its ARG
reverted, and is named on stderr and in the step summary. One unfetchable
package costs its own bump instead of the whole group's. `expat` is in that
state today: its URL template pins the `R_2_8_2` release directory, which
`${version}` does not reach, so bumping `EXPAT_VERSION` 404s. Left as a separate
data fix rather than folded in here.

## Verification

End-to-end against the real tree and the real network, not just fixtures:

```
$ sed -i 's/^ARG PCRE2_VERSION=10.47$/ARG PCRE2_VERSION=10.48/' Dockerfile
$ ./hack/refresh-source-checksums.sh --changed
pcre2 10.48: c08ae2388... -> ebcc25aad...
refreshed 1 checksum(s), 0 already current, 0 skipped
```

One line of `sources.yaml` changed, to the hash of the tarball #608 actually
pins. The skip path, on a bump that genuinely 404s:

```
$ sed -i 's/^ARG EXPAT_VERSION=2.8.2$/ARG EXPAT_VERSION=2.8.4/' Dockerfile
$ ./hack/refresh-source-checksums.sh --changed
curl: (22) The requested URL returned error: 404
warning: no URL served expat-2.8.4, leaving its checksum alone
reverted expat to 2.8.2 (2.8.4 was not fetchable)
refreshed 0 checksum(s), 0 already current, 1 skipped
$ git diff --stat Dockerfile sources.yaml    # empty
```

And a no-op on an already-current package: `refreshed 0, 1 already current`.

`hack/refresh-source-checksums_test.sh` covers the rewrite, the URL fallback,
the skip-and-revert path, `--changed` selection and the unknown-package guard
against `file://` fixtures, so it needs no network. It also asserts the
committed `sources.yaml` parses and that every `version_arg` in it resolves to
an ARG default. Runs in `check-kernel-configs` and as `make test-source-checksums`.

Each of five mutations to the script is caught by the suite: neutering the
sha256 rewrite, selecting every package in `--changed`, removing the revert
block, trying only the first URL, and dropping the unknown-package guard.

## After this merges

#601 to #608 still carry stale checksums, since they were branched before this.
Closing them lets the next nightly recreate them with the sha256 half included.


Itxaka (approved):

itxaka-agent:

Bundled with the other 2026-09-17 autobumper PRs in a single tracking view: kairos-io/kairos#4697 — one place for all bumps, changelogs, prereq (#609) and the three anomalies that need holds (automake 1.18.94 beta, shadow numeric downgrade, kernel arm64-only regen).

PR #614, opened 2026-09-17
POC: split stage1 to make kernel bumps cacheable

itxaka-agent:
> **Disclosure (rule 13)**: opened by the Kairos triage agent (itxaka-agent), operated by @itxaka. Draft, not for merge as-is — a POC to see if the CI matrix survives and to gather cache-hit numbers on a kernel bump. Every step (branch, commits, this PR body) was generated by the agent; a human will read the numbers and decide whether to invest in the full audit.

POC for kairos-io/kairos#4711.

## Idea

Today every package in the Dockerfile chains, eventually, to a single ``stage1`` layer that carries the kernel headers. Bumping the kernel invalidates ``kernel-headers-stage0`` -> ``stage1`` -> **every** downstream package cache, even the ones that never ``#include <linux/*>``. That is what makes routine kernel bumps expensive on buildx.

Split ``stage1`` into two:

Before:
```
stage1-merge --(copy skeleton)--> stage1 (has linux-headers) --> every package
                    ^
       kernel-headers-stage0 rsync into /skeleton/usr/
```

After:
```
stage1-merge --(copy skeleton, no headers)--> stage1-core --> kernel-agnostic packages
                                                   |
                                                   +--> stage1 (overlays linux-headers) --> kernel-header users
                                                            ^
                                                 kernel-headers-stage0 COPY to /usr/
```

``stage1`` is preserved as a name and its rootfs is byte-identical to what it was before, so nothing that stays on ``FROM stage1`` needs to change. The kernel-headers COPY moves out of ``stage1-merge`` and into the new ``stage1`` overlay, so ``stage1-merge`` and ``stage1-core`` no longer depend on ``kernel-headers-stage0`` at all.

## Demonstrator packages flipped to ``FROM stage1-core``

Five direct ``FROM stage1`` leaf stages, all confirmed to not ``#include <linux/*>``:

| Package       | Reason it is safe |
|---------------|-------------------|
| ``m4``        | Plain autoconf userland; libc only. Downstream ``perl`` uses only POSIX headers. |
| ``findutils`` | Plain autoconf userland; libc only. No downstream. |
| ``grep``      | Plain autoconf userland; libc only. No downstream. |
| ``gperf``     | Small perfect-hash code generator; libc only. No downstream. |
| ``flex``      | Scanner generator; libc only. No downstream. |

[...]

PR #611, opened 2026-09-17
chore(deps): consolidated 2026-09-17 dependency bumps

itxaka-agent:
> Automated triage agent (`kairos-triage-agent`) running as `@itxaka-agent`.
> This comment was generated by software. A human maintainer can override.

Single PR that supersedes the safe subset of the 2026-09-17 autobumper PRs (#601, #602, #603, #605, #606, #607, #608). Sources #609 is already merged into main and provides the checksum refresher this PR runs.

## Included bumps

| Package | Old | New | Source PR |
|---|---|---|---|
| libtirpc | 1.3.7 | 1.3.8 | #601 |
| multipath-tools | 0.15.0 | 0.15.1 | #601 |
| urcu | 0.15.6 | 0.15.7 | #601 |
| coreutils | 9.11 | 9.12 | #602 |
| systemd | 261.2 | 261.3 | #602 |
| util-linux | 2.42.2 | 2.42.3 | #602 |
| bash patches | 15 | 20 | #602 |
| curl | 8.21.0 | 8.22.0 | #603 |
| rsync | 3.4.4 | 3.5.0 | #603 |
| cryptsetup | 2.8.7 | 2.8.8 | #605 |
| openssh | 10.4p1 | 10.5p1 | #605 |
| openssl | 3.6.3 | 3.6.4 | #605 |
| kernel (linux) | 7.1.8 | 7.2.6 | #606 |
| dwarves | 1.31 | 1.32 | #606 |
| open-iscsi | 2.1.12 | 2.1.13 | #607 |
| strace | 7.1 | 7.2 | #607 |
| expat | 2.8.2 | 2.8.4 | #608 |
| libelf (elfutils) | 0.195 | 0.196 | #608 |
| libxml2 | 2.15.3 | 2.15.4 | #608 |
| pcre2 | 10.47 | 10.48 | #608 |

Kernel configs for amd64, arm64 and riscv64 are the autobumper's regenerated output from #606, applied verbatim; no arch was skipped.

`sources.yaml` sha256 fields refreshed via `hack/refresh-source-checksums.sh` for every bumped package. Expat's sha was computed manually from the fetched tarball because the refresher does not yet substitute `${version_underscore}` in url templates and expat's release tag is `R_${version_underscore}` — a follow-up to the refresher can drop that manual step.

## Excluded

| Package | Attempted | Reason |
|---|---|---|
| automake (#604) | 1.18.94 | Beta/alpha tarball, not a stable release. |
[...]

PR #615, opened 2026-09-17
POC (option B): late-bind kernel headers per consumer

itxaka-agent:
> **Disclosure (agent-authored, rule 13).** Opened by the Kairos triage
> agent running as `itxaka-agent`. Every commit here is authored and
> signed off by `itxaka-agent`; the code changes are the agent's, not
> Itxaka's personal work. Reviewer time from a human before merge is
> still expected.

Option B for kairos-io/kairos#4711. Supersedes the shape (not the
issue) of kairos-io/hadron#614, which is the POC for option A - both
PRs are meant to sit side by side so the trade-off is easy to
compare. `#614` moves a couple of things into a `stage1-core`
overlay; that surfaced the real coupling problem (bash and rsync host
several header-consumers themselves, so a lot of the tree cannot move
into a header-free `stage1-core` without also rearranging bash and
rsync). Option B here just drops the headers from `stage1-merge`
outright and re-adds them per consumer, so the base build image is
kernel-agnostic and the ~30 stage1-derived stages that do not touch a
`<linux/*>` header stop invalidating on a kernel bump.

## Commits

1. `chore(build): drop kernel headers from stage1 base` -
   `stage1-merge` no longer copies `/linux-headers` into
   `/skeleton/usr`. Consumers break in isolation and are wired back up
   in the next commit.
2. `chore(build): late-bind kernel headers in the packages that need
   them` - each header-consuming stage does
   `COPY --from=kernel-headers-stage0 /linux-headers/. /usr/include/`
   right after its `ARG` block, so only that stage's layer cache
   depends on the kernel tarball.

## Why `kernel-headers-stage0` rather than `kernel-headers`

There are two stages that expose `/linux-headers`:

- `kernel-headers-stage0` (L733 in main): `FROM make-stage0`. Only
  depends on the linux tarball + a make binary; runs
  `make headers_install` in isolation.
- `kernel-headers` (L2193 in main): `FROM kernel-base`. Runs the same
  `headers_install`, but only after the whole kernel-base graph
  (rsync, bash, gcc, openssl, elfutils, pahole, ...) is built.

Every consumer here uses the former: pulling from `kernel-headers`
would drag the whole kernel-base dependency graph into the consumer's
cache key and defeat the point of the split. The one remaining use of
`kernel-headers` (the `full-toolchain-merge` stage at L3802, which is
what people install headers *from*) is left alone because it already
sits downstream of kernel-base.

## Consumers that received the explicit COPY

| Stage | Evidence |
|---|---|
| `libcap` | `<linux/capability.h>` (task's "verify" list) |
| `util-linux` | `<linux/fs.h>`, `<linux/blkzoned.h>`, `<linux/loop.h>`, `<linux/watchdog.h>` |
| `libseccomp` | `<linux/seccomp.h>`, `<linux/audit.h>`, `<linux/filter.h>` |
| `kmod` | `<linux/module.h>` via libkmod internals |
| `libkcapi` | `<linux/if_alg.h>` (kernel crypto API) |
| `libbpf` | `<linux/bpf.h>`, `<linux/btf.h>`, `<linux/perf_event.h>` |
| `kernel-base` | scripts/mod host tooling + inherited by kernel-cloud, kernel-default, kernel-build, kernel-no-fips, kernel-fips, kernel-modules, kernel-headers, kernel-misc |
| `kbd` | `<linux/kd.h>`, `<linux/keyboard.h>`, `<linux/vt.h>` |
| `strace` | full `<linux/*>` syscall/ioctl surface |
| `libmnl` | `<linux/netlink.h>` |
| `libnftnl` | `<linux/netfilter/nf_tables.h>` |
| `iptables` | `<linux/netfilter*.h>` |
| `libnfnetlink` | `<linux/netfilter/nfnetlink.h>` |
| `libnetfilter_conntrack` | `<linux/netfilter/nf_conntrack_*.h>` |
| `libnetfilter_cttimeout` | `<linux/netfilter/nfnetlink_cttimeout.h>` |
| `libnetfilter_cthelper` | `<linux/netfilter/nfnetlink_cthelper.h>` |
| `libnetfilter_queue` | `<linux/netfilter/nfnetlink_queue.h>` |
| `lvm2` | `<linux/dm-ioctl.h>`, `<linux/fs.h>` |
| `e2fsprogs` | `<linux/fs.h>`, `<linux/fiemap.h>`, `<linux/major.h>` |
| `dosfstools` | `<linux/fs.h>`, `<linux/hdreg.h>` |
| `conntrack-tools` | `<linux/netfilter/nfnetlink*.h>` + `<linux/netfilter/nf_conntrack_*.h>` |
| `procps-ng` | `<linux/*>` for /proc parsing (sysinfo, taskstats) |
| `keyutils` | `<linux/keyctl.h>` |
| `nfs-utils` | `<linux/nfs.h>`, `<linux/nfs4.h>`, `<linux/nfs_idmap.h>` |
| `audit` | `<linux/audit.h>`, `<linux/netlink.h>` |
| `cryptsetup` | `<linux/dm-ioctl.h>`, `<linux/loop.h>`, `<linux/fs.h>` |
| `parted` | `<linux/blkpg.h>`, `<linux/hdreg.h>`, `<linux/fs.h>` |
| `tpm2-tss` | `<linux/tpm.h>` (for `/dev/tpm*`) |
| `systemd` | fanotify, keyctl, netlink, dm-ioctl, loop, bpf, seccomp, watchdog, ... |
| `lvm2-systemd` | same as `lvm2` |
| `multipath-tools` | `<linux/dm-ioctl.h>`, `<scsi/*>` |
| `openscsi` (open-iscsi) | `<linux/netlink.h>`, `<linux/genetlink.h>`, `<scsi/*>` |

32 stages total. The rest of the ~70 stage1-derived stages are now
kernel-agnostic.

## Suspects that were NOT touched

| Stage | Reason |
|---|---|
| `pam-systemd` | systemd-flavoured rebuild of pam; hook goes through libsystemd, not the kernel API |
| `shadow-systemd` | same shape as pam-systemd - libc + PAM + libsystemd, no `<linux/*>` |
| `dbus`, `dbus-systemd` | uses `<sys/*>` and glib, not kernel headers |
| `perl` | interpreter, no `<linux/*>` |
| `python-build` | interpreter, no `<linux/*>` |
| `libaio` | not a build stage in this Dockerfile (only appears as a `-download` alias); nothing to touch |

Non-suspects that stay header-free: `curl`, `openssl`, `openssl-fips`,
`expat`, `sqlite3`, `libffi`, `zlib`, `zstd`, `lz4`, `xxhash`, `attr`,
`acl`, `coreutils`, `findutils`, `grep`, `diffutils`, `gawk`, `bash`,
`readline`, `flex`, `bison`, `m4`, `autoconf`, `automake`, `libtool`,
`patch`, `binutils`, `ca-certificates`, `ncurses`, `less`, `kbd`
runtime, `xz`, `gzip`, `bc`, `pcre2`, `glib2`, `libmspack`,
`ca-certificates`, `iconv`, `libxml`, `libtirpc`, `libnl`, `libevent`,
`libucontext`, `libelf`, `argp`, `obstack`, `fts`, `elfutils`,
`urcu`, `jsonc`, `pahole` and dwarves, `pax-utils`, `cmake`, `openssh`,
`sudo` and variants, `shadow` and `shadow-base`, `pam`, `hadron-splash`,
`grub-*`, `shim`, `dracut`, `gdb-stage0`, `qemu-guest-agent`,
`open-vm-tools-build`. Any of these that turn out to need a header at
build time will get the same one-line COPY added in a follow-up.

## Measurement plan

Once CI is happy on the raw build:

1. Rebase a small kernel bump (patch-level, or even a no-op change to
   the `linux.tar.xz` sha) on top of this branch and on top of
   `main`.
2. Compare `toolchain-amd64` cache-hit percentage between the two
   builds. Target is a large jump from the 7% baseline reported in
   kairos-io/hadron#611.
3. Cross-check with `hadron-io/hadron#614` (option A) on the same
   bump so mudler can pick between the two shapes.

## Caveats

The audit is best-effort. It is a knowledge-based sweep of each
package's known role plus a check of the "may or may not" list from
the parent ticket. Two failure modes CI will catch that this PR does
not pre-empt:

- A package that only pulls `<linux/*>` through a generated header or
  an autoconf test that emits `#include <linux/*>` into a
  `config.h`-style shim. Those references do not show up in a naive
  grep of the tarball and will fail configure or compile.
- A package on the "not touched" list that turns out to consume a
  kernel header transitively via one of its libraries whose own
  headers include `<linux/*>` under `extern "C"`.

Fix in both cases is the same: add the one-line
`COPY --from=kernel-headers-stage0 /linux-headers/. /usr/include/`
right after the stage's ARG block. A heads-up commit will follow if
CI turns anything up.

## Deliberately not doing

- Not introducing a `stage1-core` overlay. Option A already covers
  that shape in kairos-io/hadron#614; this PR is the opposite
  approach so mudler has both to compare.
- Not force-pushing over `#614`. That PR stays open and untouched.
- Not touching kairos-io/kairos#4711.

Refs: kairos-io/kairos#4711
See also: kairos-io/hadron#611, kairos-io/hadron#614


itxaka-agent:

`rsync-amd64` (and `rsync-arm64`) failed at step #448 building `attr`: `libmisc/xattrat.c:21:10: fatal error: linux/xattr.h: No such file or directory`. The initial audit missed one — attr does consume kernel headers.

Pushed `16429f6` adding `COPY --from=kernel-headers-stage0 /linux-headers/. /usr/include/` to both the `attr` and `acl` stages. acl 2.4.0 ships the same `libmisc/xattrat.c` plus `<linux/openat2.h>`; it would inherit the headers from attr anyway, but the explicit COPY matches the rest of the late-bind pattern.

itxaka-agent:
> _Automated triage bot comment. Written by an LLM under itxaka's supervision — flag anything off and I'll follow up._

Pushed `b7a6041` to fix the rsync-arm64 CI failure.

`make headers_install INSTALL_HDR_PATH=/linux-headers` writes to `/linux-headers/include/linux/*.h`. The 34 per-consumer `COPY --from=kernel-headers-stage0 /linux-headers/. /usr/include/` lines therefore landed the tree at `/usr/include/include/linux/xattr.h`, and the compiler saw nothing at `/usr/include/linux/xattr.h`.

Fixed by dropping the `/include` suffix from the destination in all 34 consumer COPYs — matches what the pre-refactor `rsync /linux-headers/. /skeleton/usr/` was doing. Docker path arithmetic, ever a pleasure.

itxaka-agent:

Closing as superseded by kairos-io/hadron#617 (option D).

Same underlying flaw as kairos-io/hadron#614 (option A): the 34-consumer late-bind approach relies on an audit of "which packages need `<linux/*>`", and gnulib's `stackvma.c` includes `<linux/fs.h>` — so m4, coreutils, diffutils, gawk, findutils, grep and any other GNU userland via gnulib all need kernel headers too. That audit can't be closed. First CI red hit m4 again.

Option D (kairos-io/hadron#617) fixes it in three lines by moving the header injection to `stage1` and using `RUN --mount=type=bind` so the layer's cache key hashes the mount source's content, not the source stage's digest. Every package inherits headers as before (no audit needed) and kernel patch bumps that don't change UAPI hit cache across the whole graph.

PR #616, opened 2026-09-17
POC (option C): content-hashed kernel headers + split ARG

itxaka-agent:
> **Automated disclosure (Second Foundation rule 13):** this PR was produced by the Kairos triage agent (`itxaka-agent`) running on Itxaka's behalf. Draft, not for merge — POC to compare against options A (#614) and B (#615). Manual review before any status change.

## Intent

Option C for kairos-io/kairos#4711. Builds on top of option B (#615) — inherits the 34 late-bind consumers + the "drop headers from stage1" change — and layers two independent cache-friendliness improvements on top.

## Why

Option B fixes the "any package sees a kernel bump" problem by moving `/linux-headers` out of `stage1` and into 34 per-consumer `COPY --from=kernel-headers-stage0` sites. That works, but a `COPY --from=X` cache key is derived from *stage X's overall layer digest*, so any rebuild of `kernel-headers-stage0` still invalidates every one of the 34 consumers — even a patch-level kernel bump whose installed `linux/*.h` are byte-identical.

Option C addresses that in two independent, stackable ways:

1. **Content-hashed headers via `RUN --mount=type=bind`.** Buildkit's cache key for `RUN --mount=type=bind,from=X,source=P` uses the *content hash of the files at P inside stage X*, not X's layer digest. Replace every per-consumer `COPY --from=kernel-headers-stage0 /linux-headers/. /usr/` with `RUN --mount=type=bind,from=kernel-headers-stage0,source=/linux-headers,target=/tmp/khdr cp -a /tmp/khdr/. /usr/`. Now a rebuild of `kernel-headers-stage0` that produces byte-identical UAPI headers is transparent to every consumer.

2. **Split `KERNEL_VERSION` from `KERNEL_HEADERS_VERSION`.** New `ARG KERNEL_HEADERS_VERSION=${KERNEL_VERSION}` (default: unchanged) and a new `linux-headers-download` stage. `kernel-headers-stage0` now pulls its tarball from `linux-headers-download`; `kernel-base` and the actual kernel build stay pinned to `linux-download` at `KERNEL_VERSION`. Operators can now pin the header source separately.

## Expected wins

- Patch-level kernel bumps (7.2.6 → 7.2.7 etc.) whose `linux/*.h` are unchanged stop invalidating the 34 consumers. Mount content hash unchanged → RUN cache hits.
- UAPI-changing bumps still invalidate — unavoidable, and correct.
- Operators can pin `KERNEL_HEADERS_VERSION` to an older UAPI-stable release and let `KERNEL_VERSION` move independently for pure kernel-binary bumps, keeping every late-bind consumer cache-warm.

## Measurement plan

Same shape as option B: rebase a small kernel patch-level bump on top of this branch, rebuild, compare the `toolchain-amd64` (and rsync-amd64/arm64) cache-hit ratio against:

- #611's 7% baseline (headers baked into stage1)
- Option B's late-bind-only baseline (still invalidates all 34 on any kernel bump)

Expected for a patch-level bump with byte-identical UAPI: near-100% hit on the 34 consumers.

## Refs / supersedes

- Refs: kairos-io/kairos#4711
- Supersedes: kairos-io/hadron#614 (option A — dead, gnulib pulls linux/*.h into every GNU userland), kairos-io/hadron#615 (option B — works, but still invalidates all 34 consumers on any kernel bump)

## Risk

The biggest risk is that the buildx / buildkit version on the CI runners doesn't handle `RUN --mount=type=bind,from=<stage>,source=…` cleanly. If that turns out to be true this PR won't work and we'd fall back to option B. Otherwise no runtime behaviour change — this is a cache-key refactor, the produced layers are semantically equivalent.


itxaka-agent:

Closing as superseded by kairos-io/hadron#617 (option D).

Same underlying flaw as kairos-io/hadron#614 (option A): the 34-consumer late-bind approach relies on an audit of "which packages need `<linux/*>`", and gnulib's `stackvma.c` includes `<linux/fs.h>` — so m4, coreutils, diffutils, gawk, findutils, grep and any other GNU userland via gnulib all need kernel headers too. That audit can't be closed. First CI red hit m4 again.

Option D (kairos-io/hadron#617) fixes it in three lines by moving the header injection to `stage1` and using `RUN --mount=type=bind` so the layer's cache key hashes the mount source's content, not the source stage's digest. Every package inherits headers as before (no audit needed) and kernel patch bumps that don't change UAPI hit cache across the whole graph.

PR #618, opened 2026-09-17
POC (options E+F on D): mount trick applied to stage1-merge and full-toolchain-merge hubs

itxaka-agent:
> **Rule 13 disclosure**: opened by the itxaka-agent triage agent, not by
> a human. Not for merge as-is — this is a measurement POC. Numbers
> from CI will decide whether to split E and F into standalone
> mergeable PRs. Please leave draft.

Extend option D's mount trick to the two remaining cache hubs in the
Dockerfile — stage1-merge and full-toolchain-merge. Refs:
kairos-io/kairos#4711. Builds on: kairos-io/hadron#617.

## Option E — stage1-merge (commit `f58bb6c`)

Five `COPY --from=<stage>-stage0 /sysroot /<local>` + `RUN rsync ... /skeleton/`
pairs collapsed into one `RUN --mount=type=bind,from=<stage>-stage0,source=/sysroot,target=/tmp/<local>`
+ inline rsync. Stages transformed:

- `gcc-stage0`
- `musl-stage0`
- `busybox-stage0`
- `make-stage0`
- `binutils-stage0`

The `skeleton` copy that seeds `/skeleton` itself is left alone — no
subsequent rsync to fold in. Net diff: 10 add / 10 remove.

## Option F — full-toolchain-merge (commit `78b6f2b`)

51 intra-Dockerfile `COPY --from=<stage>` lines converted to bind
mounts: 1 skeleton base, 40 direct-into-`/merge/` package COPYs, and
10 COPY+rsync pairs (busybox, util-linux, libcap, bash, coreutils,
gcc-stage0, musl-stage0, make-stage0, binutils-stage0, kernel-headers)
collapsed into one `RUN --mount` each.

Preamble kept unchanged: the seven `COPY --link --from=` lines at the
top of the stage (rsync, attr, acl, zstd, zlib, lz4, xxhash) install
rsync and its runtime deps into the merge stage's own filesystem. A
bind mount only lives for the duration of one RUN, so those cannot be
moved to mounts without breaking every subsequent rsync invocation.

Quirks handled inline:

- `hadron-splash` is a single-file copy into `/merge/bin/hadron-splash`;
  converted via `install -Dm755` from a file-scoped bind mount.
- `kernel-misc` lands under `/merge/usr/share/kernel-misc` (not
  `/merge`); target dir pre-created with `mkdir -p`.
- `skeleton` seeds `/merge` from scratch; `mkdir -p /merge` prepended
  so rsync has a target to write into.

Net diff: 125 add / 73 remove.

## Measurement

Mount hashing has a real IO cost per RUN (buildkit walks the source
subtree to compute the content hash). The first CI run is the point
of this POC — measure whether that cost is meaningful compared to the
rebuild it avoids. Three questions the numbers should answer:

1. **Cold build cost** — does the initial from-scratch build get
   significantly slower vs option D alone (kairos-io/hadron#617) or vs
   main? If the extra hashing per RUN dominates, we back out.
2. **Unchanged rebuild** — does a rerun with no source edits fully
   cache-hit the transformed stages? It should; if it does not, the
   mount hashing is unstable and the POC fails outright.
3. **Targeted invalidation** — push a comment-only edit to one leaf
   package's stage (e.g. `attr`) and confirm that only downstream
   consumers of THAT package rebuild in full-toolchain-merge. The old
   behavior invalidates every subsequent COPY in the stage because
   the source stage digest changed; the new behavior should only
   invalidate the one RUN whose mounted content changed. If the whole
   merge stage still invalidates, F is not doing what it claims and
   we back it out separately from E.

## Kernel-misc caveat

The mount trick on the `kernel-misc` COPY does not help across kernel
bumps: kernel modules genuinely change with the kernel version, so
content-hash cache invalidation is the correct behavior there. It is
still included for consistency with the rest of the stage; documented
inline in the Dockerfile and here.

## Not for merge as-is

If the numbers land well we can propose splitting E (small, safe,
five lines of impact) and F (bigger, needs measurement of the
per-RUN mount-hashing cost) into separate mergeable PRs. Option D
(kairos-io/hadron#617) stays as the minimal proof-of-concept
regardless.


itxaka-agent:

Closing. Numbers on the first CI run show net regression vs option D alone: stage1-amd64 +53s, rsync-amd64 +51s, systemd-amd64 +1m01s, hadron-bios-amd64 (default) +4m23s. Only meaningful F win in this measurement is toolchain-arm64 (-3m12s further).

Root cause: mount source hashing has real IO cost, and its compensating win — content-hashed key hitting cache when a source stage produced byte-identical output — can't demonstrate itself in this measurement setup. Fork PR → `PR_CACHE_WRITE=false` → no PR-scoped cache writes → nothing to cache-hit on subsequent pushes. So E+F pays the hashing tax with nothing to cash it in for.

Option D (kairos-io/hadron#617) is the proven win and stays open. E+F stays viable as an idea — its architectural value shows up under main-repo push conditions where PR-scoped cache writes work AND a subsequent build produces byte-identical package outputs. Neither of those held here. Re-open a same-repo branch of E+F after D lands on main if the trade-off wants revisiting.

